PR #4, opened 2016-12-22
Make the test suite pass again

claudep:


deeplook:
This might use `requests` one day soon. Until then, great!

PR #8, opened 2016-12-23
Rewrite normaliseSvgPath and add tests

claudep:

PR #13, opened 2016-12-23
Made WikipediaFlagsTestCase pass with all files

claudep:


claudep:
Note that #8 is also required for the test suite to pass.

PR #18, opened 2016-12-23
Use ReportLab Path API for adding path operations

claudep:


deeplook:
Always great to see code getting shorter _and_ better. ;-)

PR #21, opened 2016-12-23
Stop crashing when Polylines/Polygons have odd number of coordinates

claudep:


deeplook:
Please take a breath as I'm switching to `pytest`... ;-)

PR #22, opened 2016-12-23
Use nodes

claudep:


deeplook:
Great stuff! Be warned that I'll slow-down a bit on merges...

claudep:
You've been unexpectedly quick until now, thanks. Do not feel any pressure for coming PRs :-)

PR #29, opened 2016-12-24
Add preliminary support for elliptic arcs

claudep:


deeplook:
Wasn't aware of `svg.path` (and I'm not sure its use of complex numbers is cool or freakish). Would it make sense to use more of it here?

claudep:
I was also a bit surprised by that (point -> complex number). I don't see what it really brings over a simple tuple, at least in our use case.

claudep:
We could also make use of `namedtuple` at some point, allowing to do point.x, point.y instead of point[0], point[1].

deeplook:
I would prefer named tuples, too. But if one needs the `Arc` thing, one will have to bite the _complex_ bullet (sic!), isn't it?

Also, if you add a new dependency like this one it would likely be better to replace the ugly _local import_ of `svglib` in the test suite by one that has already been installed with `python setup.py install` or `pip install svglib`. This is a bit against my thinking of being able to test something before installing it. But one could also try to use the package built with `python setup.py build`. Thoughts?

claudep:
I'd rather suggest using pip for testing, and adding `requirements.txt` (or `requirements/{base,test}.txt`).

Instead of fiddling with the path in the test files, we could create missing `__init__.py` in src and test, and then `from ..svglib import svglib`.

deeplook:
That was on my todo list, too. Just unsure if this allows to test before installing.

PR #33, opened 2016-12-24
Shapes defined in the defs section are not directly displayed

claudep:


claudep:
With this patch, all files from wikipedia symbols test dir are perfectly drawn. Yeah!

deeplook:
Wow! And the flags get better and better, too! This is a really nice present for X-mas! Thanks! ;-)

PR #36, opened 2016-12-26
Replaced prints by logging calls

claudep:


claudep:
Please review carefully the logging levels (error/warn/debug), as I was not always sure of my choices!

deeplook:
The `unusedAttrs` stuff was meant to track features of the SVG which were not (yet) handled. There is surely a better way of doing it, but I'd like to keep the idea around.

claudep:
Well... did I remove that?

deeplook:
In fact, not. I hadn't read the entire commit.

PR #41, opened 2016-12-27
Set correct x/y text positions when tspan has x/y attributes

claudep:


claudep:
Mmmh, wait, I must still check something...

claudep:
OK, I force pushed a fix. Ready now.

PR #54, opened 2017-01-03
Fixed default stroke width as per SVG specs (from 0 to 1)

claudep:


deeplook:
I don't see how the three files deleted from ``tests/NOTES.rst`` have improved in any way, have they?

claudep:
Sure, they should. Might it be the installed svglib issue on your system?

deeplook:
Argh, yes, sorry! Have to change habits...

PR #56, opened 2017-01-04
Fill unclosed subpaths when appropriate

### LINKED ISSUE #55: Paths are closed which shouldn't be
I think there might be an issue with closing paths which also shows in a series of flag examples, e.g. that of Bhutan. I've condensed the following minimal example from `tests/samples/wikipedia/symbols/Flower-of-Life-small.svg`:

```svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg
	xmlns:svg="http://www.w3.org/2000/svg"
	width="304"
	height="304"
	id="svg2"
	version="1.0">
	<g
		id="idLayerDrawing"
		style="display:inline"
		transform="translate(-348,-348)">
		<path
			style="opacity:1;fill:none;fill-opacity:1;fill-rule:nonzero;stroke:#000000;stroke-width:1;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-opacity:1;visibility:visible;display:inline;overflow:visible;enable-background:accumulate"
			id="path4481"
			d="M 543.30127,375 A 50,50 0 0 1 500,400"
			transform="translate(-43.30127,-25)"
			/>
	</g>
</svg>
```

Clearly, there is no `z` in the path here, but for some reason this seems to be ok. Maybe it's the `fill-rule:nonzero` style... (well, apparently not).

claudep:
Fixes #55.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -497,8 +497,8 @@ class SvgRenderer:
             method_name = "convert%s" % name.capitalize()
             item = getattr(self.shape_converter, method_name)(n)
             if item:
-                if name != 'text':
-                    # Style is already applied in convertText
+                if name not in ('path', 'text'):
+                    # Style is already applied in convertPath/convertText
                     self.shape_converter.applyStyleOnShape(item, n)
                 transform = n.getAttribute("transform")
                 display = n.getAttribute("display")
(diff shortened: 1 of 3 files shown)


deeplook:
Excellent, thanks!

PR #57, opened 2017-01-05
Fixed bezier shorthand path operator in corner cases

claudep:
From the specs for 's'/'S' path operator: If there is no previous command
or if the previous command was not an C, c, S or s, assume the first control
point is coincident with the current point.


deeplook:
Awesome, thanks!

PR #67, opened 2017-01-17
Replace svg.path dependency by our own elliptical arcs code

claudep:


claudep:
I eventually pushed the path.points optimization as a separate commit in that same PR.

PR #73, opened 2017-02-17
add colorConverter option

replabrobin:
Hi, when using svglib to create RL drawings for use with existing document flows it may be necessary to use specific color schemes. This pull request does 2 things:

1) use a single Svg2RlgAttributeConverter instance by passing the renderer instance to the  Svg2RlgShapeConverter instance.

2) Add optional colorConverter argument to the Svg2RlgAttributeConverter class to allow conversion to take place in convertColor.

With these in place I can use svg2rl to make CMYK etc etc drawings.

--HG--
branch : colorConverter

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -201,10 +201,10 @@ class Svg2RlgAttributeConverter(AttributeConverter):
     "A concrete SVG to RLG attribute converter."
 
     def __init__(self,color_converter=None):
-        self.color_converter = color_converter or self.identitycolor_converter
+        self.color_converter = color_converter or self.identity_color_converter
 
     @staticmethod
-    def identitycolor_converter(c):
+    def identity_color_converter(c):
         return c
 
     def convertLength(self, svgAttr, percentOf=100):
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -200,6 +200,13 @@ class AttributeConverter:
 class Svg2RlgAttributeConverter(AttributeConverter):
     "A concrete SVG to RLG attribute converter."
 
+    def __init__(self,colorConverter=None):
+        self.colorConverter = colorConverter or self.identityColorConverter
+
+    @staticmethod
+    def identityColorConverter(c):
+        return c
+
     def convertLength(self, svgAttr, percentOf=100):
         "Convert length to points."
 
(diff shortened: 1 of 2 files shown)


deeplook:
Please add one or more test cases.

svglib/svglib.py
```diff
@@ -200,6 +200,13 @@ def convertTransform(self, svgAttr):
 class Svg2RlgAttributeConverter(AttributeConverter):
     "A concrete SVG to RLG attribute converter."
 
+    def __init__(self,colorConverter=None):
+        self.colorConverter = colorConverter or self.identityColorConverter
```
claudep: We're trying to progressively depart from the old CamelCase syntax to use a more Pythonic syntax with underscores for class attribute and method names. Could you move `colorConverter` to `color_converter`, and the same for the new method?

svglib/svglib.py
```diff
@@ -1006,7 +1009,7 @@ def applyStyleOnShape(self, shape, node, only_explicit=False):
             shape.fillColor.alpha = shape.fillOpacity
 
 
-def svg2rlg(path):
+def svg2rlg(path,**kwds):
```
claudep: Please add space after comma (PEP8). A more common name is `**kwargs`.

replabrobin:
Hi Claude, I have changed to convert_color & kwargs, but I never know if I need to redo the pull or if you always pull the head etc etc.

deeplook:
I'll merge this, although I think there should be some simple test(s) producing PDF(s) with CMYK colours.

claudep:
@deeplook for a next time, could you "Squash and merge" to make that only one commit. I don't really appreciate having messages like "added Claude's pep8 space :(" in the master tree history. I hope you'll understand that. Or if my reviews are not welcome, just tell me.

deeplook:
@claudep Your reviews are very much appreciated! I'm just often using the github interface on the go on a mobile phone. And I don't think there is a squash option there, but I'll check again.

claudep:
It might be an option to activate in the project settings (under Merge button). I thought it was on by default. Of course, it's not appropriate when the different commits are purposefully separated by separation of concerns.

replabrobin:
Not exactly sure what the issue is / was, but I didn't intend any criticism of 
Claudep in the commit messages. Pep8 is a nuisance and doing one change

kwds-->kwargs

whilst forgetting the ' ' caused me to be :(

 From another project I believe it's possible to add commit hooks to prevent 
checkins with style problems although I'm not an expert (and would be against 
such in my repos).

As for grouped commits which I believe git can do I'm not sure my git is up to 
strength on that. Currently I use the git-mercurial bridge for all of my git repos.

Should probably bite the bullet and install a vim flake8 plugin or something.
--
replabrobin

PR #77, opened 2017-03-25
Render symbol definitions

### LINKED ISSUE #76: Symbol definitions are not rendered

claudep:
Fixes #76.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -401,7 +401,7 @@ class SvgRenderer:
                 parent.add(item)
         elif name == "symbol":
             item = self.renderSymbol(n)
-            # parent.add(item)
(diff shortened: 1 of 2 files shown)


claudep:
Thanks for merging quickly. What about a 0.8.1 release now?

deeplook:
Happy to do that. But might be during next week. Would it make sense to add a simple canned SVG file as an example for what can be rendered now?

claudep:
Could be. Do you mean something different than the sample files from the tests?

deeplook:
Well, maybe we could add the Asturia score if there are no copyright issues? But I kind of doubt it. Maybe @Mygod knows?

Mygod:
I suggest not to.

PR #82, opened 2017-04-09
Respect absolute coordinates for tspan

### LINKED ISSUE #81: "y" in "tspan" interpreted as relative coordinate?
I tried to convert an SVG that contains the following text node, which contains a `y` key both in the `text` tag and in the `tspan` code:

```svg
  <text
     id="text125"
     data-bb="18"
     style="font-family: &quot;Open Sans&quot;, verdana, arial, sans-serif; font-size: 12px; fill: rgb(68, 68, 68); fill-opacity: 1; visibility: inherit; white-space: pre;"
     transform="translate(24.66,0)"
     y="234"  // y is defined here...
     x="0"
     text-anchor="middle">
    <tspan
       id="tspan121"
       y="234"  // ... and here again
       x="0"
       dy="0em"
       class="line">Oct 6</tspan>
    <tspan
       id="tspan123"
       y="234"
       x="0"
       dy="1.3em"
       class="line">2013</tspan>
  </text>
```

This node disappears off the page when converting it using `svg2pdf`, presumably because in https://github.com/deeplook/svglib/blob/master/svglib/svglib.py#L720, the y coordinates are added. If I understand correctly, the inner `y` keyword should be interpreted as an absolute coordinate though, so the outer `y` should be ignored in this case.

[plot.svg.txt](https://github.com/deeplook/svglib/files/907594/plot.svg.txt)

claudep:
Fixes #81.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -691,7 +691,7 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
         fs = attrConv.findAttr(node, "font-size") or "12"
         fs = attrConv.convertLength(fs)
         for c in itertools.chain([node], node.getchildren()):
-            has_x = False
+            has_x, has_y = False, False
             dx, dy = 0, 0
             baseLineShift = 0
             if node_name(c) == 'text':
@@ -703,7 +703,7 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
                 if not text:
                     continue
                 x1, y1, dx, dy = [c.attrib.get(name, '') for name in ("x", "y", "dx", "dy")]
-                has_x = x1 != ''
+                has_x, has_y = (x1 != '', y1 != '')
                 x1, y1, dx, dy = map(attrConv.convertLength, (x1, y1, dx, dy))
                 dx0 = dx0 + dx
                 dy0 = dy0 + dy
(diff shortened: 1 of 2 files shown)


jdugge:
That looks really good, thanks for the fast response! The conversion of the plot attached to #81 does indeed work much better now, however when converting the `airbus.svg` from the `samples/misc` directory, the second line of text ("Seating layout") gets lost with this commit.

The `text`/`tspan` tag in `airbus.svg` looks like this:

```svg
<text transform="matrix(0.9387 0 0 1 312.7847 24.2568)" style="fill:#CFCFCF;stroke:none; font-size:27.9868;letter-spacing:0em;">
<tspan x="0" y="0">AIRBUS A340</tspan>
<tspan x="-10.761" y="33.487">Seating Layout</tspan>
</text>
```

There's no `y` key in the `text` span, only a `transform` matrix.

airbus

claudep:
Nice catch, the patch needs improvement then...

claudep:
I pushed a fix. I appreciate your careful review!

deeplook:
Thanks, I hope I can look at it this evening.

deeplook:
Merged. Made me look at `samples/W3C_SVG_12_TinyTestSuite/svg/text*.svg` again, and have a big sigh at the missing features here...

claudep:
Thanks, I'll try to add basic support for em units soon.

deeplook:
BTW, for @jdugge's sample file I'm getting overlapped text bits now near the origin like this: 

issue82

claudep:
Yes, that is missing em unit support.

PR #103, opened 2018-01-10
Open SVGZ output as "wb" if Python 3

JorjMcKie:


claudep:
There is indeed an issue to solve here, but here's what I would use:
```
with gzip.open(path, 'rb') as f_in, open(path[:-1], 'wb') as f_out:
     shutil.copyfileobj(f_in, f_out)
```
Inspired from Python example in https://docs.python.org/2/library/gzip.html#examples-of-usage

deeplook:
Thanks for reporting! I'd prefer @claudep's solution which looks more concise. I've tested it and will commit a path including a new test for SVGZ and a file from Wikimedia before next week. This will also be the start for including licensing terms for SVG content included in this repo.

deeplook:
This commit should address this: 7c1502af997cfba9b166238d84a72f0455916c8e. Please have a look!

PR #106, opened 2018-02-08
Don't install pytest-runner when not required on setup.py invocation

chrisbennight:
Currently every time setup.py is called (pip install, etc.) pytest-runner ends up required.  
The PR ensures inclusion only when invoked in a pytest context.

The current behaviour caused some awkwardness when included as a transitive dependency.

Pattern is from the pytest-runner docs, ref:
https://pypi.python.org/pypi/pytest-runner
(see "Considerations->Conditional Requirements")

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,9 @@ v = sys.version_info
 if (v.major, v.minor) < (2, 7):
     install_requires.append('argparse')
 
+needs_pytest = {'pytest', 'test', 'ptr'}.intersection(sys.argv)
+pytest_runner = ['pytest-runner'] if needs_pytest else []
+
 setup(
     name='svglib',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -24,7 +27,7 @@
     keywords='SVG, PDF, reportlab, conversion, graphics',
     url='https://github.com/deeplook/svglib',
     install_requires=install_requires,
-    setup_requires=['pytest-runner'],
+    setup_requires=[] + pytest_runner,
```
chrisbennight: left the [] in to signal a place where traditional setup_requires libs could be added - but it's also entirely redundant here

deeplook:
Thanks!

PR #110, opened 2018-03-06
Exclude a W3C SVG test file that freezes svglib

claudep:


claudep:
Of course, it would be better to debug the issue, but I don't have the time right now.

claudep:
@deeplook How much effort should we continue to put on the Python 2 version? Just to know if it's worth taking the time to fix the Python 2 test failure.

deeplook:
My gut feeling would be to support Python 2 as long as it is supported itself, but it would be ok to simply skip one test for this version using pytest's skip feature, and maybe revisit it later.

claudep:
Whoo-hoo!

deeplook:
Nice! Thanks! 💯

Issue #111, opened 2018-03-08
Generated PDF is all black


jsvachon2:
Hi,

I am using pygal to generate a SVG file which displays just fine in my browser. But when I convert it to any other format (PDF, jpg, png) using svglib, all I am getting is a picture completely black.

I am on OSX, do I need any additional package/library installed?

Thanks

claudep:
You may imagine that without the SVG file, we cannot say much about this :-)

jsvachon2:
Yeah I guess this would help. The tool does not support svg so I renamed it to a text file

[cpuusage.txt](https://github.com/deeplook/svglib/files/1793929/cpuusage.txt)

jsvachon2:
Not sure if this will help but using the svg2pdf binary is also generating a black  image...

/tmp $ svg2pdf cpuusage.svg

[cpuusage.pdf](https://github.com/deeplook/svglib/files/1794780/cpuusage.pdf)

deeplook:
@jsvachon2 Not sure, I've heard of http://pygal.org/en/stable before. Your example seems to be created with v. 2.4.0. Do you observe that effect for all your samples created with pygal or only some, or only this one? :)

claudep:
The reason is that style is defined by a stylesheet in your SVG document, and svglib doesn't support that yet (just opened #112 to track that issue).

claudep:
@jsvachon2 Could you please also attach the expected output as image?

PR #116, opened 2018-03-10
Added support for CSS style sheets in SVGs

claudep:
Use tinycss2 and cssselect2 to read and apply style sheets.
Refs #111.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,7 +64,8 @@ Features
 Known limitations
 -----------------
 
-- stylesheets are not supported (only the style attribute)
+- support for stylesheets is still experimental. Please report any
+  bug or shortcoming on the `svglib issue tracker`_.
 - clipping is limited to single paths, no mask support
 - color gradients are not supported
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,4 @@
 reportlab
 lxml
+tinycss2==0.6.1
+cssselect2==0.2.1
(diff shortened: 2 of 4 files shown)


claudep:
This may need a bit more real-world testing before integration in master.

deeplook:
I can only check this very superficially. It looks ok to me, but it would be nice, maybe, to have just a bit more test cases, and feedback from other users. Can we identify any sample files inside the official SVG test-suite we're running with CSS that work better than before? 

And for a release like 0.9.0 we should then at least mention the feature in the current README, so people know CSS support is there in principle. And then we have something to expand later, ideally in some real, Sphinx-based documentation, say for 1.0, and on readthedocs.io.

claudep:
I may have a little time to spend on it tomorrow.

claudep:
I didn't find any stylesheet usage in the official SVG test suite.
I updated the limitation note in the README file. I'm afraid that feedback from users won't come until we release that code on PyPi.

README.rst
```diff
@@ -64,7 +64,8 @@ Features
 Known limitations
 -----------------
 
-- stylesheets are not supported (only the style attribute)
+- experimental support for stylesheets was introduced in 0.9.0. Please report
```
deeplook: I would skip any version number here until we get a release out.

deeplook:
@claudep See my comment on the release number above. Otherwise fine, and let's merge it!

PR #122, opened 2018-03-28
allow for a list of fonts

replabrobin:
I added a class method Svg2RlgAttributeConverter.listAttrSplit for splitting list attributes and changed convertFontFamily to allow alternate font names.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -53,7 +53,7 @@ __date__ = '2017-04-22'
 
 XML_NS = 'http://www.w3.org/XML/1998/namespace'
 STANDARD_FONT_NAMES = (
-    'Times', 'Times-Roman', 'Times-Italic', 'Times-Bold', 'Times-BoldItalic',
+    'Times-Roman', 'Times-Italic', 'Times-Bold', 'Times-BoldItalic',
     'Helvetica', 'Helvetica-Oblique', 'Helvetica-Bold', 'Helvetica-BoldOblique',
     'Courier', 'Courier-Oblique', 'Courier-Bold', 'Courier-BoldOblique',
     'Symbol', 'ZapfDingbats',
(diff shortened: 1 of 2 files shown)


claudep:
Excellent, I have extracted the `find_font` function separation in an other commit and pushed it in master (cef7f11). Could you rebase your patch? It should be easier to review now.

replabrobin:
Not sure exactly why the tests are failing. They work for me in windows; I am trying the tests in linux and will commit when they work there as well.

claudep:
I suspect that's because fontconfig will still return a font name even if you feed it a weird and unexistent font name.

replabrobin:
Yes bilbo returns a default with fc. I think then that the idea of allowing a list of fonts is completely stupid. As it means that the first item will always produce a match. A later test insists that SomeFont will return a font if font config is available. Surely this strategy is wrong; additionally the default font returned by FC will vary depending on which fonts are available. Perhaps the strategy should be to return an exact match if available and then return the bad match. If that sounds reasonable then I'll implement that.  On windows I would expect style="font-family:bilbo arial" to return arial, but font search is not well implemented by reportlab.

claudep:
> Perhaps the strategy should be to return an exact match if available and then return the bad match. If that sounds reasonable then I'll implement that.

+1

replabrobin:
I hate git. The rebase has failed miserably. I will delete my repo and start again. Git sucks compared to mercurial (at least for me). I think it was wrong to start the merge without full understanding. To implement a sensible strategy will require possibly two loops.

replabrobin:
OK I have pushed changes to my repo implementing a two search strategy. First search is exact; if that fails then we search inexact.

replabrobin:
This PR cannot succeed as there are changes that clash after your checkin.  Can you just drop the PR and I will make another. My code is now passing tests on windows and linux.

claudep:
Please look at #123.

PR #118, opened 2018-03-15
Fixes #117 - Prevented ZeroDivisionError in elliptical arcs

claudep:

PR #123, opened 2018-03-28
 Allow list of fonts in font family property

claudep:
Here's a slightly different approach of the PR #122.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -59,7 +59,7 @@ STANDARD_FONT_NAMES = (
     'Symbol', 'ZapfDingbats',
 )
 DEFAULT_FONT_NAME = "Helvetica"
-_registered_fonts = set()
+_registered_fonts = {}
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


svglib/svglib.py
```diff
@@ -53,27 +53,33 @@
...
 
 logger = logging.getLogger(__name__)
 
 Box = namedtuple('Box', ['x', 'y', 'width', 'height'])
 
 
 def find_font(font_name):
-    if font_name in STANDARD_FONT_NAMES or font_name in _registered_fonts:
-        return font_name
+    """Return the font and a boolean indicating if the match is exact."""
```
deeplook: "Boolean" or "bool" (sorry for nit-picking :)

claudep: Thanks, I learned something today :smiley:

replabrobin:
Hi Claude, this works for me and my windows test do pass.

claudep:
Thanks Robin for initiating this feature!

replabrobin:
I have updated my svglib repo to address issue 117 (the ellipse zerodiv error), pr 114. I needed the latter as part of code to do backfilling of clip-paths as I had an example where the def of the clip path was at the end of the file but was defining a rect.

How should I go about presenting this code to you? The clip-path backfill seems to work well.

claudep:
You may review #114 and #118 so as we can merge them in master soon. I'm not sure about what code you want to present, is this svglib code? Feel free to point us to some branch of yours with this code.

replabrobin:
This is pure svglib; I have an example where the clip-path is used several times as 
<path clip-path="url(#p38b2f8081d)" d="M -54.609375 320.4...../>

and later we see 
<defs>
  <clipPath id="p38b2f8081d">
   <rect height="277.2" width="697.5" x="112.5" y="43.2"/>
  </clipPath>
</defs>

this fails to clip in latest svglib; so I have made changes to make this work.

The result is in master https://github.com/replabrobin/svglib

deeplook:
👍  Thanks, @replabrobin and @claudep!

Issue #124, opened 2018-04-03
PDF Export with Embedded SVG


jdgwf:
I've created the following test SVG (I've simplified this ugly example immensely after discovering this issue): 

**SVG Preview Screenshot**:
![screenshot from 2018-04-03 10-09-15](https://user-images.githubusercontent.com/316824/38258132-4e1b47ac-3727-11e8-8402-5e5646217be3.png)

**PDF Render**:
![screenshot from 2018-04-03 10-09-37](https://user-images.githubusercontent.com/316824/38258152-5762d0be-3727-11e8-93e3-a39650bc9465.png)

Will paste my (again, ugly) test SVG code below.

**What appears to be happening**:
svglib is trying to use one of the embedded SVG dimensions as the parent document dimension.

jdgwf:
Here's the svg as a txt file:
[drawing.svg.txt](https://github.com/deeplook/svglib/files/1872119/drawing.svg.txt)

claudep:
Yes, it's known that embedding SVGs is currently buggy (this was revealed by the `Slovenia.svg` flag output in the current test suite).

jdgwf:
I apologize for the duplicate ticket then. Any suggestions where to look to help squash this bug? I'll be mulling over the code to help patch this :)

claudep:
No need to apologize, there was no properly-named ticket about this until now.
The problem is probably around the `renderSVG` method, which assumes it is always the top-most node of the file.
https://github.com/deeplook/svglib/blob/master/svglib/svglib.py#L526

jdgwf:
Ok I have a working Alpha quality patch - testing now
![screenshot from 2018-04-03 10-46-45](https://user-images.githubusercontent.com/316824/38260111-5d7c94f8-372c-11e8-9a51-c8b59f9b6744.png)

jdgwf:
Looks like I may have to open a pull request once I'm more sure ;)

claudep:
Show us the code, dude :-)

jdgwf:
Scaling needs to be fixed...

jdgwf:
Here are the two functions I've applied code to class SvgRenderer:

    def __init__(self, path=None,color_converter=None):
        self.attrConverter = Svg2RlgAttributeConverter(color_converter=color_converter)
        self.shape_converter = Svg2RlgShapeConverter(path,self.attrConverter)
        self.handled_shapes = self.shape_converter.get_handled_shapes()
        self.definitions = {}
[...]

PR #129, opened 2018-04-06
Fixed #121 So the viewbox is scaled

### LINKED ISSUE #121: When using viewBox attribute, canvas size is not scaled to width and height of svg attribute
When trying to convert the following svg to png

```
from reportlab.graphics import renderPDF, renderPM

svgtemplate="""
<svg   width="1250" height="1250"  viewBox="-40 -40 80 80">
<rect width="30" height="30" style="fill:red" />
</svg>
"""
def png(svg,pngfilename):
    text_file = open("template.svg", "w")
    text_file.write(svg)
    text_file.close()
    drawing = svg2rlg("template.svg")
    renderPM.drawToFile(drawing, pngfilename)

png(svgtemplate,'Test_viewBox.png')
```

I receive a png that is only 80px x 80px.
It should be 1250 x 1250 and the rect should be scaled accordingly.

(svglib 0.8.1 python 3.6.4 Windows 7)

tomturner:


.gitignore
```diff
@@ -14,5 +14,8 @@ docs/_build
 *.swp
 *.egg
 env/
+venv/
 .workon
 dist/
+.idea/
+.DS_Store
```
claudep: @deeplook may have a different opinion, but as for me, those user-specific or platform-specific things should be excluded in users global gitignores (I'd prefer removing some of existing lines than continue to add more).

svglib/svglib.py
```diff
@@ -284,9 +284,11 @@ def convertLength(self, svgAttr, percentOf=100, em_base=12):
             return float(text[:-2]) * em_base
         elif text.endswith("px"):
             return float(text[:-2])
+        elif text.endswith("mm"):
+            return float(text[:-2]) * mm
```
claudep: Nice improvement, but this should happen in a separate pull request, with a test. Would you be able to do that please?

svglib/svglib.py
```diff
@@ -425,15 +427,18 @@ def __init__(self, path=None,color_converter=None):
...
         for xlink in self.waiting_use_nodes.keys():
             logger.debug("Ignoring unavailable object width ID '%s'." % xlink)
-
-        main_group.scale(1, -1)
+        scale_x = self.width / self.box.width
+        scale_y = self.height / self.box.height
+        main_group.scale(scale_x, -scale_y)
         main_group.translate(0 - self.box.x, -self.box.height - self.box.y)
-        drawing = Drawing(self.box.width, self.box.height)
+        drawing = Drawing(self.width, self.height)
```
claudep: We are about to commit support for embedded SVGs in #124, which will heavily touch this part of the code. It would be great to wait for that other PR and then try to rebase yours onto the new code.

svglib/svglib.py
```diff
@@ -801,7 +808,7 @@ def convertPath(self, node):
         subpath_start = []
         lastop = ''
 
-        for i in xrange(0, len(normPath), 2):
+        for i in range(0, len(normPath), 2):
```
claudep: It's indeed strange that `xrange` usage doesn't crash in current tests, as it's not present by default in Python 3. I guess reportlab imports are doing some magics to alias it. I'm in favor of the change, but this should also be part of another PR with the other usage of `xrange` in svglib.

.gitignore
tomturner: Ok far enough. The fist one was PYCharm the second was a file that get generated on a Mac

svglib/svglib.py
tomturner: Done in #130

svglib/svglib.py
tomturner: Let me know when you want me to do this.

svglib/svglib.py
tomturner: Done in #131

deeplook: I've commented there about some simple test code that would make this nicely complete. Otherwise thanks!

.gitignore
deeplook: Yep, these should surely not go into the repo.

svglib/svglib.py
claudep: #124 has been merged, so you can try to rebase your patch.

claudep: Oh, and if you have an example (SVG + expected result), that would be great.

tomturner: Ok I will write the patch again and send it in

svglib/svglib.py
deeplook: Sorry, but is this still relevant? I can't figure out why it got stalled for so long.

claudep:
As the original ticket is now closed, I guess this PR can be closed also.

PR #131, opened 2018-04-07
Added millimetre support

tomturner:


deeplook:
@tomturner Thanks for this first contribution of yours! Could you think of a simple test to be added, maybe to `TestLengthAttrConverter` in `test_basic.py`?

tomturner:
Just found out that mm support was already in the library by default. I have added a comment and a test. What do you think?

PR #134, opened 2018-04-10
Scale svg node viewbox relative to width/height

claudep:
Refs #121.


deeplook:
@claudep If this is this still relevant, can we merge?

PR #143, opened 2018-06-06
Change license classifier from GPL to LGPL3

lordmauve:
I noticed that although the license is LGPL the trove classifier is still GPL. This updates the classifier to the correct string for LGPL-licensed packages.


deeplook:
Thanks!

PR #144, opened 2018-07-08
cast the rgb value to float

winglq:
The result of int(val)/100.0 will be 0.01 or 0. As int(val) will do the cast first. If val<1 then int(val)=0, otherwise if val == 1 then the result is 0.01

Signed-off-by: Liu Qing <winglq@gmail.com>


claudep:
Thanks! Do you think you would be able to suggest a regression test also?

winglq:
Hi claudep,
I think a regression test may cover the following scenario.

```
Svg2RlgAttributeConverter.convertColor("fill:rgb(100%,98.399353%,94.099426%)") should return Color{
r:1, g:0.98399, b:0.94099,a:1} instead of Color{r:1,g:0,b:0,a:1}
```
I am not familiar with the code structure now, I could add a test when I get some free time.

PR #145, opened 2018-07-13
Added CHANGELOG content for the 0.8.2 release

claudep:


claudep:
Hey @deeplook, here's my suggestions for the CHANGELOG. How about a release soon?

deeplook:
This would justify 0.9.0 label, and even more so if we could check off some remaining PR. I might have the time over the next few days for making such a release. I would probably go with a 0.9.0rc1, first, just to test my rusty knowledge about the process. I've also discovered https://github.com/kennethreitz/setup.py only yesterday, but I guess something like that will be only when reaching 1.0.

deeplook:
Fine, please add under 0.9.0 to come!

claudep:
@deeplook What about putting latest changes first, reversing ordering?

deeplook:
Sure, if that better meets the norm...

claudep:
Pushed in 2a05d508f90ca9abcbf4cbf9824f133ce5f17292 and reordering in 2b16e395572a63dc895ca3d8555a2d1b98464864

Issue #140, opened 2018-05-10
Make a new release 0.9.0, RSN


deeplook:
There have been many requests for a new release of svglib on PyPI, which I have procrastinated for way too long (my sincerest apologies!). I've collected some thoughts in https://github.com/deeplook/svglib/wiki/Release-Process, but in the end it will likely happen rather by mimicking a few techniques I found to be successfully used when instantiating projects with https://github.com/audreyr/cookiecutter/ using a template like https://github.com/audreyr/cookiecutter-pypackage which I've used for something different, before. It relies on a couple of tools like `bumpversion` (not really maintained anymore) and `twine`, but using anything else will be better than what there is now (not much).

claudep:
I'm happily using `twine` for some of my projects.

claudep:
I wonder if it would be worth shipping #116 (CSS support) in that release, even if it's not polished (better than no support at all?).

deeplook:
I would put in everything new since 0.8.1, as simple as that.

claudep:
Well, PR #116 is not merged yet, that's why I'm asking.

deeplook:
I won't have any time to test this in any other way than reviewing code and test cases, and build results from travis-ci. Just looked briefly now. Will comment in https://github.com/deeplook/svglib/issues/116.

MinchinWeb:
I would add a request for a new PyPI version. There's a small error in version 0.8.1 (and 0.8.0) where `xrange` [is called](https://github.com/deeplook/svglib/blob/v0.8.1/svglib/svglib.py#L747), but that function is not available on Python 3. So right now, Python 3 compatibility is broken. But the this has been fixed on GitHub.

Thanks for the library!

Overdrivr:
Would you like some help for automating the release of the library ?
Something I rely extensively on is configuring appveyor/travis to deploy only on tags.
This way, the only thing you need to perform a release is to create a new tag in Github.

deeplook:
@Overdrivr We have appveyor and travis set up for svglib. As you offer help that might accelerate things. I would like to keep appveyor and travis checking every commit, but if we can have releases with every new tag that would be fine. If you have more info to point me to I will be glad to have a look!

Overdrivr:
Both things are not exclusive, on the contrary. Basically the idea is to run the same build & test suite, whether it's a simple commit or a commit with tag. And simply on tags, go the extra step and push to pypi. 

I'll open a PR then, it will give you a better idea.

[...]

PR #151, opened 2018-08-13
Update svglib.py

skidzo:
line 747, in convertPath'
for i in xrange(0, len(normPath), 2):', "NameError: name 'xrange' is not defined"

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -50,6 +50,11 @@ logger = logging.getLogger(__name__)
 
 Box = namedtuple('Box', ['x', 'y', 'width', 'height'])
 
+# python3??
+try:
+    xrange
+except NameError:
+    xrange = range
 
 class NoStrokePath(Path):
     """


claudep:
Thanks but this was already fixed in 6eb609f.

claudep:
@deeplook we really need a release :-)

deeplook:
@claudep I know, I know. I put some hope on https://github.com/deeplook/svglib/issues/140, which might be the fastest solution...

claudep:
Well, isn't a release essentially running two commands?:
`python setup.py sdist`
`twine upload dist/svglib-x.y.z.tar.gz`

deeplook:
@claudep There are files to synchronize manually with the latest version number (at least the README), and much output from the test suite that is actually not "tested", but ideally needs to be "eye-balled". This has taken me considerable time in the past. But I'm willing to spend parts of the weekend to come on making (and releasing) 0.9.0b (and ignore some quality concerns) which should serve as a point of departure for fixing some of the remaining issues in 0.9.0 one or two weeks later (I won't be able to do much on that myself). When that happens I think it would be ok to nuke 0.9.0b from PyPI, again (showing an appropriate warning in the README). Thoughts?

claudep:
Ah, of course, manually checking rendered files take time. Hopefully regressions should be limited, as we have a bit more tests now. Your plan sounds good.

skidzo:
good luck with the release and the weekend ;-)

PR #160, opened 2018-09-06
Added svgstring2rlg

woolyninja:
Allows converting of svg strings to reportlab graphics instead of requiring a file. This comes in handy if getting an SVG string from an outside API, etc.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -1230,6 +1230,25 @@ def svg2rlg(path, **kwargs):
     return drawing
 
 
+def svgstring2rlg(svgstring, **kwargs):
+    "Convert an SVG file to an RLG Drawing object."
+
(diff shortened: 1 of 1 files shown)


claudep:
Thanks for the patch. I would be tempted to adapt `svg2rlg` to also accept a string in its first parameter. @deeplook, any opinion?
Anyway, this will need a test.

woolyninja:
Yeah, that probably makes more sense. I'll make that change tonight and add some tests.

woolyninja:
Looking at the existing tests made me realize I could just wrap my string with io.StringIO to make it act like a file:
rlg = svg2rlg(io.StringIO(svgstring.decode('utf-8')))

So I don't think there's any need for this PR. I'm going to close it.

PR #162, opened 2018-09-26
Fix Python 3 compatibility issues

bayangan1991:


claudep:
Thanks for the suggestion, but you must be using an old version of svglib, as the `xrange` and `STRING_TYPES` symbols are no longer in use in the most recent svglib code.

bayangan1991:
Ok thanks. I installed straight from pip though. Could that version be updated?

claudep:
There is a beta version available, try installing it with `pip install --pre ...`.

bayangan1991:
Thank you!

Issue #173, opened 2019-02-24
passing file-like object to svg2rlg


Akuli:
I want to use `svg2rlg` with `io.BytesIO`, but it wants a path to a file, and not a file-like object, so I need to create a `tempfile.NamedTemporaryFile`, copy the `BytesIO` content there, and then `svg2rlg` with that temporary file's name. This is a bit inconvenient, and can be slow with large files. Ideally `svg2rlg` could also take a file-like object instead of a path.

Currently `svg2rlg` has `svgz` code that creates an `svg` file in the same directory as where the `svgz` file is. That must be changed to implement this because `BytesIO` objects don't even have a directory where they are, but I think that using a temporary file or `BytesIO` would be better anyway.

claudep:
There is support for file-like objects, the test suite is using `StringIO` extensively as input. It might chokes on `BytesIO` as it expects text input. What's the error you get?

Akuli:
I didn't even try it because `svg2rlg` does `os.remove(path)` at the end. There's no way that could possibly work with file objects.

Akuli:
I read the code a bit more carefully. The `os.remove(path)` is for `svgz` files only anyway, but the `path` is passed to `SvgRenderer`. Then it passes it to `Svg2RlgShapeConverter`, which inherits its `__init__` from `SvgShapeConverter` that sets it to `self.svg_source_file`. The `Svg2RlgShapeConverter` then has code like `os.path.join(os.path.dirname(self.svg_source_file), xlink_href)` which would break if `svg_source_file` was a file object.

I'm feeling like if I pass a file object to `svg2rlg`, it might "work", until I try to use some feature that just doesn't happen to work with file objects (or my code gets an SVG file that has some stuff that svglib handles differently).

Akuli:
Reading the code even more: the `os.path.dirname(self.svg_source_file)` code runs only if the SVG contains images that are not embedded in it as base64, so I think it'll be fine for my needs. Might be nice to document this though. Sorry for the wall of text!

Which encoding should I use if I have an SVG file's content as bytes, and I want to create a `StringIO`-like file object from it? `utf-8`?

deeplook:
If you got up to here, you might want to contribute a patch I guess it would be well accepted.

On the encoding, I'd always start with UTF-8.

claudep:
Thanks for the explorations. I just added a fix/test in #174 to prevent a nasty crash in a case you mention.

Akuli:
[...]

Issue #97, opened 2017-10-14
Support *pathlike* objects.


MinchinWeb:
What am I proposing is that `svglib.svglib.svg2rlg()` accept a path-object created by `pathlib.Path()` (added to the standard library in Python 3.5) rather than just strings.

I have pathlib is a wonderful addition to my Python programming, and it would be a touch simpler if I could use it directly with svglib.

deeplook:
I like the idea. I just have never used `pathlib` myself, yet. It should be a simple thing to contribute, which I'm encouraging you to do (while not giving up Python 2.7 compatibility).

MinchinWeb:
Thanks @deeplook ; I'll look into it, but I can't give you a timeline at the moment!

PR #174, opened 2019-02-25
Refs #173 - Gracefully fail when resolving relative xlink in file-like inputs

claudep:


Akuli:
```
            if not isinstance(self.svg_source_file, str):
                logger.error(
                    "Unable to resolve image path '%s' as the SVG source is not a file system path." % xlink_href
                )
                return None
```

Doesn't this fail if the source file is a pathlib Path object, because those are not strings? I don't know whether path objects are supported elsewhere, but if they are, might be nice to change this to something else.

There are other things that use `isinstance` and `str` with paths, too.

claudep:
Support for `Path` objects is asked in issue #97. It is not yet supported.

deeplook:
Thanks!

Issue #175, opened 2019-03-03
External svg image does not work


metro69:
I am trying to convert a simple svg that references another svg (see below).

Using svglib 0.9.0, reportlab 3.5.13.

test.svg is correctly rendered by Chrome.
svg2pdf returns with:
`Unable to read the image another.svg#rectangle. Skipping...`

Metro

test.svg:

`<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"    width="1000" height="1000">
    <image xlink:href="another.svg#rectangle" class="blue"/>
</svg>
`

another.svg
`<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"    width="1000" height="1000">
    <rect x="558.5" y="570" width="75" height="50" id="rectangle" />
</svg>`

claudep:
You are right, this is not supported yet. First, support for external svg should be added, then targeting a specific element in an linked svg should be implemented (probably a bit harder).

claudep:
The pull request above is implementing the first stage of the fix only.

metro69:
Thanks. I can now include svg's into another svg by removing the '#rectangle' and render to pdf using svg2pdf.

However, the transform does not seem to work yet. For instance, the svg below puts both rectangles at the same location.

I assume that is part of the next phase fix. Is there something I can do in the interim?

Metro

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"    width="1000" height="1000">
<g>
    <image xlink:href="another.svg" class="blue"/>
    <image transform="translate(200,200)" xlink:href="another.svg" class="blue"/>
    </g>
</svg>
```

claudep:
#179 should improve the situation, we are not far from being able to target an external fragment. Could you test it on your use case?

metro69:
The last update seems like a step back. I uninstalled svglib and reinstalled directly from your repo:
`pip install git+https://github.com/claudep/svglib@xlink_href
`
None of the two rectangles show up anymore in the generated pdf.

Metro

claudep:
Thanks for trying, I will explore that in the next days.

claudep:
[...]

PR #176, opened 2019-03-03
Refs #175 - Add initial support for embedded external svg

claudep:


deeplook:
Great stuff, thanks! That feature was missing for a long time.

PR #178, opened 2019-03-07
Fixing deprecation warning for raw string

Rogalek:
test of my application code shows me that:
/python_modules/svglib/svglib/utils.py:19
  /python_modules/svglib/svglib/utils.py:19: DeprecationWarning: invalid escape sequence \d
    floats = [float(seq) for seq in re.findall('(-?\d*\.?\d*(?:e[+-]\d+)?)', value) if seq]

Fixing that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 .coverage
-.idea
 MANIFEST
 coverage.xml
 nosetests.xml
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .coverage
+.idea
 MANIFEST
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 .coverage
+.idea
```
claudep: I'm not sure this should be part of the patch, this typically belongs to your global ignore file.

Rogalek: Yes, you are right. I removed that from PR.

claudep:
Thanks!

PR #179, opened 2019-03-10
Refs #175 - Cleaned xlink_href handling

claudep:
We now also apply image attributes when target is a SVG file.


deeplook:
@claudep Feel free to merge, if it's ok... Thx!

PR #183, opened 2019-03-21
Fix bug where temporary files are not closed

ajbrock:
I noticed that while mass-converting a bunch of SVGs to PNGs (400+) this script was opening hundreds of temporary files, not closing them, and crashing with an OSerror of "too many files open." The source stems from the  `tempfile.mkstemp` actually opening the temporary file it creates (it's a separate open from the `open(path, 'wb')`, and not closing it (guess garbage collection doesn't grab it properly?) This fix resolves the issue for me.


deeplook:
LGTM, thanks!

PR #184, opened 2019-03-21
Respect multi-coordinate tspan

ajbrock:
This commit adds the ability to handle <tspan> elements where multiple x or y coordinates are provided (as in [this example](https://drive.google.com/file/d/1CEjdn02y1hYrCFZK0aNRLRQKepRUphB6/view?usp=sharing), which the svglib does not currently render correctly but chrome handles just fine). This just modifies the map(convertLength) call (which presently ignores all but the first coordinate in such <tspan> elements) to take into account all of the tspan coordinates, and then to construct the resulting shape one character at a time.

Sidenote, I'm not really sure what PR etiquette is for this project, I'm just PRing changes I've made to my own copy that fix issues I'm running into, please let me know if I need to change anything =)

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -968,7 +968,14 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
                     continue
                 x1, y1, dx, dy = [c.attrib.get(name, '') for name in ("x", "y", "dx", "dy")]
                 has_x, has_y = (x1 != '', y1 != '')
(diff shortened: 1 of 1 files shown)


claudep:
Thanks! It would be great if you could add a test for that new functionality.

deeplook:
Please go ahead! I wanted to add a PR twmplate for ao long now. GitHub has feature for that...

claudep:
Thanks for the patch, however I rewrote it in #219.

PR #185, opened 2019-03-23
Add ability to render fragments of external SVG files

claudep:


claudep:
It's not perfect yet, but I think this should add basic support for rendering external svg fragments.

claudep:
Refs #175.

Issue #83, opened 2017-04-10
Remaining little flag artefacts


deeplook:
Three flags in `samples/wikipedia/flags` show minor but clearly visible artefacts with paths and/or clipping: `Kyrgyzstan.svg`, `Slovenia.svg` and `The_republic_of_china.svg`.

claudep:
After a quick look, these are probably different issues.
For `The_republic_of_china.svg`, it's probably an issue with `viewbox` vs svg height and width.
For `Slovenia.svg`, it's a svg inside another svg (currently unsupported).
For `Kyrgyzstan.svg`, it's probably an issue with `clipPath` being a circle (should check if Reportlab supports that or not).

deeplook:
Likely correct...

claudep:
Issue #124 should address the SVG embedding problem.

claudep:
I think all previously reported artefacts are now gone.

PR #189, opened 2019-05-17
Update utils.py to handle mishandled circular arcs

ajbrock:
There are some cases where svglib will incorrectly render circular arcs. For example, [this svg](https://pastebin.com/3a4aF601) should look approximately like [this](https://i.imgur.com/Ru95Anz.png) when rendered to a raster image, but will instead ends up with [one of the arcs messed up](https://i.imgur.com/0HGIm9m.png). This appears to stem from the `r > 1` case of `end_point_to_center_parameters` when r is close to 1, resulting in the r value being kept at 1 (rather than `1 / r -1` , which should be approximately 0). This pull fixes these render errors; it may be more robust to do something like check again if the radicand is negative or just take `max(1 / r - 1, 0)` at some point, but this works for all of the SVGs I'm using this library for.

I'm not sure what tests to write for this (or the previous PR) but I'm going ahead and submitting this one as it does solve my problem.


deeplook:
I'm fine with this but I don't have the time to run a detailed test to see what changes for the test suite. One might expect some things to improve most likely for the flags examples. There should be a list of flags showing some issues right now. Will add a reference later...

claudep:
The flag issue is #83, but I don't think that arc rendering has anything to do with those. Of course, we should still check this doesn't add any regression in flags.

claudep:
@ajbrock Are we authorized to integrate your sample svg in the svglib test samples?

ajbrock:
Yes, please feel free to include my sample svg!

deeplook:
@ajbrock Thanks! Maybe you can add it directly to this PR to some appropriate place, like `tests/samples/misc`?

ajbrock:
Added to `tests/samples/misc`--the CC3.0 license mentioned in there is fine by me for this svg, but please let me know if you'd like me to add anything else to `LICENSES.rst`!

deeplook:
@ajbrock My initial idea for that `LICENCES.rst` file was to list source and license for each example. At the moment there is only one mentioned in there. As you can see this is not finalised, yet... But, ideally you would add a little section for the example you've contributed. The formatting might be changed again a little bit at some later point.

claudep:
Thanks Andy, pushed in 9184153.

PR #198, opened 2019-07-01
Fix AppVeyor build

### LINKED ISSUE #197: Decide on continued Windows support
At the moment we have AppVeyor fails for every merge. I don't see anybody except @nicoddemus who would care for this and be able to solve it. If I'm right maybe it's worth doing something about it, and keep some records of how to do it? Otherwise I feel we should drop official Windows support with the next release. Opinions?

The last error from AppVeyor was this: https://ci.appveyor.com/project/deeplook/svglib/builds/25657065

nicoddemus:
Fix #197

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,9 +6,10 @@ environment:
   - TOXENV: "py37"
 
 install:
-  - C:\Python36\python -m pip install tox
+  - C:\Python37\python -m pip install -U pip
+  - C:\Python37\python -m pip install -U tox
 
 build: false  # Not a C# project
 
 test_script:
-  - C:\Python36\scripts\tox
+  - C:\Python37\python -m tox


nicoddemus:
@deeplook ready for review. 👍 

I'm watching the repository to keep a watch over the Windows builds, but if any of them fails for reasons which are not code related please don't hesitate in pinging me.

deeplook:
That would switch from Python 3.6 only to 3.7 only, right? Just nosey: is there a reason for not testing both?

nicoddemus:
> That would switch from Python 3.6 only to 3.7 only, right? Just nosey: is there a reason for not testing both?

No, the interpreter invoking tox is not necessarily the same which will be created by tox and be used to run the commands. This can be verified by the pytest banner:

**py27**

```
============================= test session starts =============================
platform win32 -- Python 2.7.16, pytest-4.6.4, py-1.8.0, pluggy-0.12.0
```

**py35**

```
============================= test session starts =============================
platform win32 -- Python 3.5.4, pytest-5.0.0, py-1.8.0, pluggy-0.12.0
```

And so on.

That's why on Linux it is perfectly OK and common to execute `python3 -m tox -e py27` to execute the tests under Python 2, even though you are executing tox through Python 3.

PR #201, opened 2019-07-06
Feature/97 support pathlike objects

girstenbrei:
Hi svglib devs!

I wanted to do something, so I started implementing #97.
Please, have a look at my attempt and feedback me! Is the python3-check ok like this? Are the tests ok using a tempdir? Is the documentation good? Is #173 affecting this?

Greets,
Chris


deeplook:
@girstenbrei Do you want to continue on this from the perspective of svglib having dropped support for Python 2?

claudep:
Superseded by #261

PR #210, opened 2019-09-30
Rename svg2pdf to svg2pdf.py

softwareCobbler:
The 'extensionless' file is probably a linuxism that I'm not aware of. As a Windows user I had to go in and manually set this file's extension to `.py` in order to use it after installing from pip.
I am in no way pushing hard for this change, I just felt this was the most appropriate place to ask about it (didn't want to file an 'issue' or whatever). If this will be a "won't fix," would you be willing to spare a few minutes to educate me on the use-case of an extensionless script?
File extension grumbling aside - it works well, so thanks for building / maintaining this.


deeplook:
Extensionless scripts may look unusual for Windows users, where all executables traditionally have extensions like `.exe` or `.msi` etc. (I haven't used Windows for very long, except the few times some company would pay me to do so).

For Python-centric projects/repos I'm quite sure this is not so unusual. But it would be interesting to see some proper stats derived from repository hosteres like GitHub, GitLab and BitBucket. I'd say it maybe depends on the size of the project. If there is a lot of functionality spread over a more or less large number of modules, and there are a couple of tools/commands with some CLI I think few people would be surprised to see something "extensionless" as a CLI-based perspective. Think `git` etc. Something much smaller (and more Pythonic) I've used today again is `asciinema`. If you have a small repo with essentially only one small module that can be also used from the command-line nobody would be surprised to see the extension appear in the CLI command, too.

In which category the `svg2pdf` "tool" inside `svglib` would fit most is maybe highly debatable. In the many (sort of) years of its existence nobody has brought up this as an issue so far. And I have recently not contributed much myself to it, but have gladly welcomed @claudep more or less taking over the project maintenance.

Maybe the bottom line for me would be: there was nobody with strong feelings in favour of Windows contributing much development time to this project at any point of time so far with the honorable exception of @replabrobin, say. We actually declared not to support Windows at all for a long time, and it is only recently that @nicoddemus kindly started to provide AppVeyor support.

So while I don't want to "break" existing `svglib` users' habbits with a simple rename it might be very easy to come up with something like an additional extensionless alias for `svg2pdf` for Windows users. If you modify your PR along similar lines, we might be closer to a deal. ;)

softwareCobbler:
Thank you for taking the time to reply, Deeplook. On Windows we've got the 'pathext' environment variable that allows flagging extensions as executable. I'm not certain if there's any way to add 'no extension' to that list.
Anyway, best I've got: I wrote a little Windows shim to just curry the command line arguments into the real svg2pdf.
Thanks, I appreciate the git training.

PR #224, opened 2020-01-30
extract labels

revesansparole:
This merge request to allow svg2rlg tools to preserve the label of groups (used by inkscape as layers)

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -50,57 +50,57 @@ class TestPaths:
 
         mapping = (
             ("",
-                []),
+             []),
 
             ("M10 20, L 30 40 ",
-                ["M", [10, 20], "L", [30, 40]]),
+             ["M", [10, 20], "L", [30, 40]]),
 
             ("M10 20, L 40 40Z",
-                ["M", [10, 20], "L", [40, 40], "Z", []]),
+             ["M", [10, 20], "L", [40, 40], "Z", []]),
 
             ("M10 20, L 30 40 40 40Z",
-                ["M", [10, 20], "L", [30, 40], "L", [40, 40], "Z", []]),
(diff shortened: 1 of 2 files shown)


deeplook:
Thanks! I'm fine with this if you provide a test case as well.

revesansparole:
I added three simple tests. Please tell me if they match your criteria.

tests/test_basic.py
```diff
@@ -428,6 +428,73 @@ def test_css_stylesheet(self):
...
+                   style="opacity:1;fill:none;fill-opacity:1;stroke:#ff0000;stroke-width:3;stroke-linecap:butt;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;marker-end:url(#Arrow1Send)"
+                   d="M 601.06712,388.63166 H 954.67961"
+                   id="path819" />
+                </g>
+            </svg>
+        ''')))
+        main_group = drawing.contents[0]
+        gr, = main_group.contents
+        assert gr.label == "x_axis"
+
```
claudep: Thanks for the tests. I know testing one thing at a time is good practice, but honestly in this case, I think we could as well put all assertions in the same test, as they are closely related and the source svg is identical.

claudep:
Could you explain somewhere (comment in the code/test or commit message) what advantage we get by preserving ids and labels?

tests/test_basic.py
revesansparole: I like the idea of keeping atomic tests so I kept the three functions but I changed the svg code to test different cases

PancakeFriday:
I'm looking for this feature as I want to extract specific SVG layers from an SVG file. Is this going to get merged (only asking since there has been no change since February).

claudep:
Thanks, I squashed and removed formatting changes in #258.

PR #235, opened 2020-03-07
Apply flake8 checks to source code

claudep:


claudep:
@deeplook Is it OK for you?

deeplook:
Sure, even using black would be fine.

claudep:
black is not my friend right now, we'll see…

PR #239, opened 2020-03-28
Fixes #238 - Prevent stroke-ing clipping paths

### LINKED ISSUE #238: SVG with clipPath forces stroked border
Thanks for publishing this library, it's very helpful for including matplotlib plots in PDF reports.

I've run into trouble when my SVG needs to be clipped to its bounding rectangle. I'd like to just clip it, but it also draws the the bounding rectangle.

Here's a small example of what I'm trying to do. The circle is drawn half outside the top boundary of the SVG, so I want to see a semicircle.

    import svgwrite
    from io import BytesIO
    
    from reportlab.platypus import SimpleDocTemplate
    
    
    def go():
        doc = SimpleDocTemplate("example.pdf")
        story = []
        svg_drawing = svgwrite.Drawing(size=("400px", "200px"))
        clip_path = svg_drawing.defs.add(svg_drawing.clipPath(id='border_clip'))
        clip_path.add(svg_drawing.rect(size=(400, 200), fill='red', stroke='blue'))
        svg_drawing.add(svg_drawing.circle((200, 0),
                                           60,
                                           clip_path='url(#border_clip)'))
        svg_text = svg_drawing.tostring()
        svg_bytes = svg_text.encode()
        pdf_drawing = svg2rlg(BytesIO(svg_bytes))
    
        story.append(pdf_drawing)
        doc.build(story)
    
    
    go()

The PDF looks like this:

Screenshot from 2020-03-28 14-31-38

Here's what I want the PDF to look like, without drawing the bounding rectangle:

Screenshot from 2020-03-28 14-48-16

### Analysis
I did some investigation, and the problem appears to be that [`Canvas.clipPath()`] is receiving `stroke=1`. That tells it to stroke the clip path. It's hard for me to tell exactly how that stroke value is decided, but it seems to be left as the default when calling [`convertRect()`] on the rectangle in the clipping path. Perhaps it should copy the stroke from the rectangle instead of using the default. As you can see in my snippet, I set the stroke, but it was ignored.

### Workaround
It's ugly, but I worked around this by monkeypatching `Canvas.clipPath()`. That just changes it from always drawing the clip path to never drawing it. Some users might want to choose the stroke for each clip path.

Here's my example with the monkeypatch. It produces the second PDF, above.

    import svgwrite
    from io import BytesIO
    from reportlab.pdfgen.canvas import Canvas
    
    from reportlab.platypus import SimpleDocTemplate
    
    original_clip_path = Canvas.clipPath
    
    
    # noinspection PyUnusedLocal,PyPep8Naming
    def patched_clip_path(canvas,
                          aPath,
                          stroke=1,
                          fill=0,
                          fillMode=None):
        stroke = 0
        # noinspection PyArgumentList
        return original_clip_path(canvas, aPath, stroke, fill, fillMode)
    
    
    def go():
        Canvas.clipPath = patched_clip_path
        doc = SimpleDocTemplate("example.pdf")
        story = []
        svg_drawing = svgwrite.Drawing(size=("400px", "200px"))
        clip_path = svg_drawing.defs.add(svg_drawing.clipPath(id='border_clip'))
        clip_path.add(svg_drawing.rect(size=(400, 200), fill='red', stroke='blue'))
        svg_drawing.add(svg_drawing.circle((200, 0),
                                           60,
                                           clip_path='url(#border_clip)'))
        svg_text = svg_drawing.tostring()
        svg_bytes = svg_text.encode()
        pdf_drawing = svg2rlg(BytesIO(svg_bytes))
    
        story.append(pdf_drawing)
        doc.build(story)
    
    
    go()

I'm using these versions:

* svglib 1.0.0
* reportlab 3.5.42
* Python 3.8.2
* Ubuntu 18.04

[`convertRect()`]: https://github.com/deeplook/svglib/blob/v1.0.0/svglib/svglib.py#L981-L985
[`Canvas.clipPath()`]: https://github.com/MrBitBucket/reportlab-mirror/blob/0191da4e0c8a1421517f385548dcfca5e3a2bf63/src/reportlab/pdfgen/canvas.py#L1767

claudep:
Thanks Don Kirkby for the report.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,11 @@
 ChangeLog
 =========
 
+Unreleased
+----------
+
+- avoid stroking clipping paths (#238)
+
 1.0.0 (2020-03-22)
 ------------------
 
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -142,6 +142,8 @@ class ClippingPath(Path):
         props = Path.getProperties(self, *args, **kwargs)
         if 'fillColor' in props:
             props['fillColor'] = None
+        if 'strokeColor' in props:
+            props['strokeColor'] = None
         return props
 
 
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -172,7 +172,8 @@ class TestPaths:
                  xmlns:xlink="http://www.w3.org/1999/xlink" width="660" height="480">
                 <defs>
                     <path id="my-clipping-path" d="M99,176 L 110 170 112 172Z"/>
-                    <rect id="my-clipping-rect" x="88.155" y="163" width="419.69" height="20.68"/>
+                    <rect id="my-clipping-rect" x="88.155" y="163" width="419.69" height="20.68"
+                          fill="red" stroke="blue"/>
                 </defs>
                 <clipPath id="my-clip-path">
                     <use xlink:href="#my-clipping-path"/>
@@ -186,7 +187,11 @@ class TestPaths:
         ''')))
         assert isinstance(drawing.contents[0].contents[0].contents[0], svglib.ClippingPath)
         # Clipping rect was converted to a path
-        assert isinstance(drawing.contents[0].contents[1].contents[0], svglib.ClippingPath)
+        rect_clip = drawing.contents[0].contents[1].contents[0]
+        assert isinstance(rect_clip, svglib.ClippingPath)
+        # Fill and stroke properties are force-deleted.
+        assert rect_clip.getProperties()['fillColor'] is None
+        assert rect_clip.getProperties()['strokeColor'] is None
 
 
 def force_cmyk(rgb):


claudep:
@donkirkby, what do you think of this patch, could you test it?

donkirkby:
Yes, that fixes it for me. Thanks for the quick patch!

Is it possible to tell the diagram whether or not to stroke its clipping rectangle, or are you better to just draw a rectangle if you want one?

claudep:
From the 1.1 SVG specs: `The raw geometry of each child element exclusive of rendering properties such as ‘fill’, ‘stroke’, ‘stroke-width’ within a ‘clipPath’ conceptually defines a 1-bit mask (with the possible exception of anti-aliasing along the edge of the geometry) which represents the silhouette of the graphics associated with that element. `
So yes, if you want to draw the clipping perimeter. I think you have to define a non-clipping rectangle.

PR #260, opened 2021-01-13
Added fontmap and some tests for it.

morphonedev:
As mentioned, this would be a possible solution to handle font-weight , font-style mapping from svg to certain reportlab fonts.
Since the main goal is to provide a mapping from svg font-family, font-weight and font-style combinations to reportlab fontnames, I think this is better placed in svglib and not reportlab.


.gitignore
```diff
@@ -16,3 +16,4 @@ docs/_build
 env/
 .workon
 dist/
+tests/samples
```
claudep, line 19: This is wrong, there are real and valid files in that folder.

svglib/svglib.py
```diff
@@ -47,7 +47,7 @@
     normalise_svg_path,
 )
 
-__version__ = '1.0.1'
+__version__ = '1.0.2'
```
claudep, line 50: Please don't touch the version number in this patch.

svglib/svglib.py
```diff
@@ -64,60 +64,188 @@
...
 
 logger = logging.getLogger(__name__)
 
 Box = namedtuple('Box', ['x', 'y', 'width', 'height'])
 
 split_whitespace = re.compile(r'[^ \t\r\n\f]+').findall
 
-
-def register_font(font_name, font_path):
+class FontMap:
```
claudep, line 76: I would vote for pushing `FontMap` and font-related code to a separate `fonts.py` file.

tests/test_fonts.py
```diff
@@ -0,0 +1,239 @@
...
+@pytest.mark.parametrize("fontname", list(STANDARD_FONT_NAMES)
+    )
+def test_find_font_defaults(fontname):    
+    name, exact = find_font(fontname)
+    assert name == fontname
+    assert exact == True
+
+
+def test_plain_text():
+    drawing = svg2rlg(io.StringIO(textwrap.dedent(u'''\
```
claudep, line 59: Please remove the `u` prefixes which are now unneeded.

tests/test_fonts.py
```diff
@@ -0,0 +1,239 @@
...
+        <?xml version="1.0"?>
+        <svg width="777" height="267" xml:space="preserve">
+            <text style="fill:#000000; stroke:none; font-size:28;">
+            <tspan style="font-family:'Courier';font-style:italic;font-weight:bold;">TITLE    1</tspan>
+            <tspan x="-10.761" y="33.487">Subtitle</tspan>
+            </text>
+        </svg>
+    ''')))
+    main_group = drawing.contents[0]
+    assert main_group.contents[0].contents[1].fontName == 'Courier-BoldOblique'
```
claudep, line 239: It's a good idea to create a dedicated test file for font-related stuff. But please move also here font-related tests from `test_basic.py`.

claudep (changes requested):
A strong patch! See my comments for some improvements to do.

svglib/svglib.py
```diff
@@ -64,60 +64,188 @@
...
+            else:
+                result_name += weight.lower().capitalize()
+        if style != 'normal':
+            result_name += style.lower().capitalize()
+        return result_name
+
+    @staticmethod
+    def guess_font_filename(basename, weight='normal', style='normal', extension='ttf'):
+        """
+        Try to guess the actual font filename depening on family, weight an style,
```
deeplook, line 111: See typos in `depening` and `an`.

svglib/svglib.py
```diff
@@ -64,60 +64,188 @@
...
+        """
+        self._map = {} # for faster searching we use internal keys for finding the matching font
+
+        self.register_default_fonts()
+
+    @staticmethod
+    def build_internal_name(family, weight='normal', style='normal'):
+        """
+        If the weight or style is given, append the capitalized weight and style to the font name.
+        E.g. family="Arial", weight="bold" and style="italic" then the internal name would be "Arial-BoldItalic", this mimics the default fonts naming shema
```
deeplook, line 94: See typo in `shema`.

svglib/svglib.py
```diff
@@ -64,60 +64,188 @@
...
+        return result_name
+
+    @staticmethod
+    def guess_font_filename(basename, weight='normal', style='normal', extension='ttf'):
+        """
+        Try to guess the actual font filename depening on family, weight an style, 
+        this works at least for windows on the "default" fonts like, Arial, courier, Times New Roman etc.
+        """
+        prefix = ''
+        if weight.lower() == 'bold' and style.lower() != 'italic':
```
deeplook, line 115: Maybe shorten code by assigning `weight.lower()` and `style.lower()` to temporary vars, and use those.

svglib/svglib.py
```diff
@@ -64,60 +64,188 @@
...
+            prefix = 'i'
+        filename = f'{basename}{prefix}.{extension}'
+        return filename
+
+    @staticmethod
+    def use_fontconfig(font_name, weight='normal', style='normal'):
+        NOT_FOUND = (None, False)
+        # Searching with Fontconfig
         try:
+            pipe = subprocess.Popen(
```
deeplook, line 129: What if Fontconfig is not available?

tests/test_fonts.py
```diff
@@ -0,0 +1,239 @@
+import pytest
```
deeplook, line 1: For import blocks like this I recommend following the best practice of grouping it into built-in, third-party and project dependencis, in this case: (io, textwrap), (lxml), (svglib). Maybe that's even in PEP8, but there are tools to autoformat code like this, like black and isort. I hope the code follows this in other places, already.

.gitignore
morphonedev: My bad, this was just a local change, I didnt intend to commit.

svglib/svglib.py
morphonedev: I just wanted to indicate a change, but sure I can keep it at 1.0.1

deeplook:
@morphonedev This looks really great! I know I haven't used fonts extensively myself, but I'm surprised nobody else had a need to "complain" or "fix" it like this before. This is a great contribution. Thanks!!

I've added some more or less minor comments, and agree with @claudep otherwise. A separate file like fonts.py is recommended as we already have 1541 lines of code in svglib.py. And after this PR maybe we can find a way to refactor it even a bit more.

svglib/svglib.py
morphonedev: This would probably break the backward compatibility with previous versions. If this is ok for you I can move the code to a separate file.

claudep: As the long as "public" symbols are imported in this file, I don't see what would be the compatibility issue. If there are, then we can document them in the release notes.

svglib/svglib.py
morphonedev: Would not the following except blocks caputre this case and return NOT_FOUND ?

tests/test_fonts.py
morphonedev: Should the u then also be removed from the test_basic.py ?

claudep: Absolutely, but in a separate PR to not mix concerns.

deeplook: Yes, we don't support Python 2 anymore.

tests/test_fonts.py
morphonedev: Do you mean moving the whole class TestTextNode: to the test_fonts.py?

morphonedev:
I am so sorry i screwed up the squashing and somehow killed the last commit. I had to recommit the changes for the original pull request.

.gitignore
claudep: Pushed in 1e94905cfe, thanks for the suggestion!

svglib/fonts.py
```diff
@@ -0,0 +1,199 @@
...
+            registerFont(TTFont(font_name, font_path))
+        except OSError:
+            return NOT_FOUND 
+        except TTFError:
+            return NOT_FOUND
+        # Fontconfig may return a default font totally unrelated with font_name
+        exact = font_name.lower() in os.path.basename(font_path).lower()
+        return font_name, exact
+
+    def register_default_fonts(self):
```
claudep, line 96: I don 't know the exact cost of registering fonts, but do you think you can find some lazy method to register fonts only when needed? What do you think?

svglib/fonts.py
```diff
@@ -0,0 +1,199 @@
...
+            reg_name, exact = self.register_font(font_name, guessed_filename)
+            if reg_name is not None:
+                return reg_name, exact
+            else:
+                return FontMap.use_fontconfig(font_name, weight, style)
+
+
+_font_map = FontMap() # the global font map
+
+def register_font(font_name, font_path=None, weight='normal', style='normal', rlgFontName=None):
```
claudep, line 187: Are those methods only for keeping backwards compatibility? If yes, then we should add comments explaining that, and show deprecation warnings when they are run.

svglib/fonts.py
```diff
@@ -0,0 +1,199 @@
...
+            # Try first to register the font if it exists as ttf
+            guessed_filename = FontMap.guess_font_filename(font_name, weight, style)
+            reg_name, exact = self.register_font(font_name, guessed_filename)
+            if reg_name is not None:
+                return reg_name, exact
+            else:
+                return FontMap.use_fontconfig(font_name, weight, style)
+
+
+_font_map = FontMap() # the global font map
```
claudep, line 185: If this symbol is meant to be imported (as it's the case in svglib.py), I wouldn't start it with an underscore. The underscore convention is meant for private stuff AFAIK.

svglib/fonts.py
morphonedev: I thought about 2 usecases.
First one: use one font map for all conversions. I think this is the default use. In this case they get stored inside the global font_map and those functions you mentioned can be used.
Second one: The user wants a specific font map for one conversion or a set of conversions. This can be achieved by creating a FontMap instance and then passing it to SvgRenderer object as constructor parameter.

svglib/fonts.py
morphonedev: In the current solution the result of register_font can be used to determine if the font was successfull registered. Moving the actual registration in reportlab to a later point would also move the possible errors to the actual conversion process. I personally think that handling those error during the earliest possible time would be better.

svglib/fonts.py
morphonedev: Yeah you are right. The is just an artifact from moving the font related stuff into a separate fille.

deeplook:
I'm fine with it if you address the unused imports reported by LGTM above, too.

svglib/fonts.py
claudep: I don't agree. If some font has issues but is not used in the document to convert, it is pointless to raise an error. But anyway, we can also optimize that later in a follow-up patch.

morphonedev:
Regarding the unused import alerts. I am not sure how to get around them, if you want to keep backward compatibility. I dont know any way to export the register_font and find_font methods from svglib without importing them beforehand. This is regarding the point : https://github.com/deeplook/svglib/pull/260#discussion_r559039642. If backward compatibility is not needed I can simply remove them and the user has to import them from svglib.fonts.

deeplook:
Re the imports... ok, I see now. You could make some dummy use of the imported symbols to please LGTM, if only by assigning thwm to some private vars. With a one öine comment of saying why this makes sense.

svglib/fonts.py
morphonedev: I simply thought, that currently a user would just register those fonts which are actually needed. If he uses the global font map for all conversions the font would only be registered once anyway. The default fonts dont register any font in reportlab at all, since those are just mappings from svg fonts to one of the fifteen standard reportlab fonts.  But you are right this can be addressed in a separate path.

claudep:
Would the `# NOQA` comment satisfy LGTM? Those dummy lines are bad looking for me :-(

morphonedev:
Since the #NOQA unfortunately did not work, how about something like:
```
# To keep backward compatibility, since those functions where previously part of the svglib module
from .fonts import (register_font as fonts_register_font, find_font as fonts_find_font)  # noqa

def register_font(font_name, font_path=None, weight='normal', style='normal', rlgFontName=None):
    return fonts_register_font(font_name, font_path, weight, style, rlgFontName)

def find_font(font_name, weight='normal', style='normal'):
    return fonts_find_font(font_name, weight, style)
```

deeplook:
@morphonedev Maybe even better with leading underscores indicating they should not be used?

```python
from .fonts import (
    register_font as _fonts_register_font,
    find_font as _fonts_find_font
)  # noqa
```

deeplook:
Anytging missing for getting this merged?

morphonedev:
From my side there are no open points anymore. If I missed something go ahead and please point it out. 
@deeplook and @claudep thank you for maintaining this great project. As you can propbably tell I am pretty new to this whole github contribution thing. In fact this was my first. So thank you for putting this much effort in to help an newbie improving.

claudep:
Thanks, I'll make a final round of review before pushing it.

claudep:
Thanks again, pushed in 83d2ec043e0

PR #266, opened 2021-01-17
Remove u string prefixes

morphonedev:
Remove u''', since its not longer necessary in python 3.x

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -218,7 +218,7 @@ class TestPaths:
         assert group is None
 
     def test_clipped_path(self):
-        drawing = svglib.svg2rlg(io.StringIO(textwrap.dedent(u'''\
+        drawing = svglib.svg2rlg(io.StringIO(textwrap.dedent('''\
             <?xml version="1.0"?>
             <svg version="1.1" xmlns="http://www.w3.org/2000/svg"
                  xmlns:xlink="http://www.w3.org/1999/xlink" width="660" height="480">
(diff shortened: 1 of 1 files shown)


claudep:
Please provide two different commits, one for imports and another one for the `u` prefix.

tests/test_basic.py
```diff
@@ -13,20 +13,17 @@
...
-from lxml import etree
 from tempfile import NamedTemporaryFile
 
-from reportlab.graphics.shapes import (
-    _CLOSEPATH, _CURVETO, _LINETO, _MOVETO, Group, Path, Polygon, PolyLine, Rect,
-)
+import pytest
+from lxml import etree
+from reportlab.graphics.shapes import (_CLOSEPATH, _CURVETO, _LINETO, _MOVETO,
+                                       Group, Path, Polygon, PolyLine, Rect)
```
claudep, line 21: I don't see the point in changing the indenting method here.

morphonedev: It was not about the intentation but more about the import grouping as suggested here by deeplook: https://github.com/deeplook/svglib/pull/260#discussion_r559038684. I simply used the `Sort import` function of visual studio code. But I reverted it.

claudep:
Fine, can you please squash the commits and set a commit message more appropriate to the changes?

And of course, the import sort stuff is valuable too (for another pull request), I just questioned the reformating of the shapes import line, which was not needed.

PR #270, opened 2021-02-03
Add streamlit demo

deeplook:


claudep:
Did you really meant to commit that .mov video? I'm a bit worried on two things:

 - distribution: do we want to distribute a video when people install svglib? I fear it's a bit wasted space for most people.
 - the `.mov` format is not the best supported, cross-platform-wise.
 
What about uploading that video on some video hosting infrastructure instead, and then link to it somewhere, maybe in the README?

deeplook:
I understand your concerns, especially when adding more demos like this in the future. OTOH, it's "distributed" only when you clone the repo, not when you pip install it, and it's 1 Mb only. I've sticked with .mov (generated by Streamlit) because GitHub renders it when linked from a MD/ReST page, and has better controls than a GIF (mainly full screen and audio). GIFs also get pretty big the larger the image dimensions are and the more the content changes. But if you know of a reliable, dedicated external service, that doesn't spam readers with ads, please let me know.

claudep:
You could try to host it on a PeerTube instance, e.g. https://peertube.linuxrocks.online.

PR #274, opened 2021-02-04
Modernize code to Python 3.6+

eumiro:
- Use f-strings
- Compact dict creation
- Use OSError as the general system-related exception
- Fix logging formatting
- Use builtin to open files


deeplook:
Nice! Looks like a bot should be able to do that, one would hope.

eumiro:
There's some stuff you can do with `pyupgrade`, but not everything, the rest was search&replace. I wanted to pytestize your tests, but for that I needed to convert path manipulation to `pathlib.Path` and for that I needed to convert string formatting to f-strings. So first step done :-)

PR #276, opened 2021-02-07
Cache tests/samples/wikipedia/{flags,symbols} for CI

### LINKED ISSUE #273: Cache some external files downloaded during test runs
To avoid connectivity issues during running CI tests and save some bandwidth, it should be great if we could cache access to the SVG test suite and the Wikipedia flags. https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows has some tips for that.

@eumiro, do you think that would be feasible?

eumiro:
Does this even work?

Closes #273

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,13 @@ jobs:
         with:
           path: "tests/samples/wikipedia/symbols"
           key: "wikipedia-symbols-${{ matrix.python-version }}-${{ matrix.os }}"
+      - name: "Cache for w3c svg12 tinytestsuite"
+        uses: actions/cache@v2
+        env:
+          cache-name: "cache-w3c-svg12-tinytestsuite"
+        with:
+          path: "tests/samples/W3C_SVG_12_TinyTestSuite"
+          key: "w3c-svg12-tinytestsuite-${{ matrix.python-version }}-${{ matrix.os }}"
       - uses: "actions/setup-python@v2"
         with:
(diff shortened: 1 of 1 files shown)


eumiro:
It looks like it works:

```
Run actions/cache@v2
  with:
    path: tests/samples/wikipedia/flags
    key: wikipedia-flags-3.7-ubuntu-latest
  env:
    cache-name: cache-wikipedia-flags
Received 2501875 of 2501875 (100.0%), 25.7 MBs/sec
Cache Size: ~2 MB (2501875 B)
/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/dc749133-b5b9-4158-8b64-d934215ddbf0/cache.tzst -P -C /home/runner/work/svglib/svglib
Cache restored from key: wikipedia-flags-3.7-ubuntu-latest
```

claudep:
That's great. Do you think we should also cache the `W3C_SVG_12_TinyTestSuite.tar.gz` file?

eumiro:
Yes, I've added the directory in a new commit (you can squash them in the merge process). If it works, the next step could be to refactor and put the whole cacheable stuff into a single directory with subdirectories to make the `ci.yml` file smaller.

eumiro:
And did you see that the three `pypy3` tests “do run but do not sow”? They do not throw an error, but nothing happens in the test step.

deeplook:
Let's check for pypy3 separately. Else nice, although I guess we save more bandwidtb than time here.

claudep:
:+1:

PR #278, opened 2021-02-12
Update utils.py

lueis:
Allow Capital E in E notation-floats


deeplook:
Would `[eE]` instead of `e` not also do it?

claudep:
Also, a test would be nice.

claudep:
Superseded by #282

PR #279, opened 2021-03-11
introduce svg2pdf.1 man page

MicroJoe:
It is always nice to have a man page for executables.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,7 @@ setup(
     packages=['svglib'],
     py_modules=[],
     scripts=['scripts/svg2pdf'],
+    data_files = [('share/man/man1', ['svg2pdf.1'])],
     classifiers=[
         # see http://pypi.python.org/pypi?%3Aaction=list_classifiers
         'Development Status :: 4 - Beta',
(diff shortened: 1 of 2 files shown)


claudep (approved):
Nice, thanks!

deeplook:
Yay! How do we install this with pip/conda?

MicroJoe:
> Yay! How do we install this with pip/conda?

Added a new commit for `setup.py` to automatically install the man page on the system.

deeplook:
I don't know how much sense it would make to add a test for this, maybe like below? @claudep?

```python
import sys
import subprocess

import pytest

@pytest.mark.skipif(sys.platform.startswith("win"), reason="Man pages not available on Windows.")
def test_installed_manpage():
    """Test if svg2pdf man page was installed."""
    try:
        subprocess.check_output(["man", "svg2pdf"])
        assert True
    except subprocess.CalledProcessError:
        assert False
```

MicroJoe:
I think it does not make sense to test this. You want to be able to run all of the tests before installing the package, but this test would verify that the manpage is installed *after* the package is installed. So it would always fail because tests should be ran before installing the package. In the end I think this kind of test is not necessary. You want to test the code, not the installation process.

My two cents.

claudep:
I can live without a test for that.

deeplook:
I guess you're right. ;)

PR #295, opened 2021-08-26
Migrating to setup.cfg

### LINKED ISSUE #294: Use a setup.cfg
Discussions about the future PR using setup.cfg

victorbnl:
Closes #294

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,3 @@
-#!/usr/bin/env python3
-# -*- coding: utf-8 -*-
-
 import sys
 from setuptools import setup
 
--- a/setup.py
+++ b/setup.py
@@ -12,8 +12,5 @@ pytest_runner = ['pytest-runner'] if needs_pytest else []
 
 setup(
     install_requires=install_requires,
-    setup_requires=[]+pytest_runner,
-    data_files=[
-        ('share/man/man1', ['svg2pdf.1'])
-    ]
+    setup_requires=[]+pytest_runner
 )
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,58 +1,16 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
claudep, line 2: With Python 3, this line should not be needed.

setup.py
```diff
@@ -1,58 +1,16 @@
...
-        'Programming Language :: Python :: 3.9',
-        'Programming Language :: Python :: 3 :: Only',
-        'Topic :: Documentation',
-        'Topic :: Utilities',
-        'Topic :: Printing',
-        'Topic :: Multimedia :: Graphics :: Graphics Conversion',
-        'Topic :: Software Development :: Libraries :: Python Modules',
-        'Topic :: Text Processing :: Markup :: XML',
-    ],
+    setup_requires=[]+pytest_runner
```
claudep, line 15: Please add spaces around `+`.

claudep (commented):
Thanks, small changes required (and rebasing would be nice).

victorbnl:
Did you mean rebasing for not having these two commits above?

claudep:
I asked the rebase because I committed a tweak to be able to know which flag is failing to download in the test suite.

claudep:
@deeplook  was right about the Afghanistan flag! Anyway, this has nothing to do with this patch, we can go ahead.

deeplook:
Using better tooling like in https://github.com/deeplook/svglib/pull/296 will make this type of contributions hopefully superfluous. ;)

PR #296, opened 2021-08-27
Add/apply initial pre-commit hooks

deeplook:
Adding more hooks and dev dependencies later.


deeplook:
Implemented in a very minimal initial version with https://github.com/deeplook/svglib/pull/318.

PR #304, opened 2021-11-27
Update README.rst

asears:
Fix broken link to reportlab


deeplook:
@replabrobin Can you confirm that http://reportlab.org is gone forever?

deeplook:
@replabrobin I can see there is content under reportlab.org, although it's not quite what I'd expect. 😉

deeplook:
@replabrobin I can see there is content under reportlab.org, although it's not quite what I'd expect. 😉 Can you elaborate?

claudep:
If the domain name is lost and taken by another entity, I guess this patch should be committed ASAP.

claudep (approved):

PR #307, opened 2021-12-24
Factorized fetching network files in test suite

claudep:


deeplook:
Nice, thanks! It's a long time I actually thought to use the `requests` package because it does simplify a few things, but this is already a good refactoring!

claudep:
Sure I thought about requests too, but in my opinion, our limited usage does not really justify adding a new dependency.

PR #308, opened 2022-01-19
Override the default fillOpacity and strokeOpacity by the respective color property

ChrisBuergi:
Override the fillOpacity and strokeOpacity if the defaults for those properties are set and fillColor or strokeColor has an alpha value (opacity) set.

If this is not done, the default opacity (1) will be used regardless of
the alpha values of the corresponding colors.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -1338,6 +1338,12 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
             else:
                 logger.debug("Ignoring transform: %s %s", op, values)
 
+    def safeApplyShapeAttr(self, shape, rlgAttr, value):
+        try:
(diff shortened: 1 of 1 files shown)


claudep:
Thanks Chris, I slightly changed the patch (and added tests) to not need the new `applied_default_value` dict: #310
Is that new patch looking good for you? Also, I would like to know about the `safeApplyShapeAttr` of your patch, against what value are you guarding there (would be great to complement the tests with those values)?

ChrisBuergi:
Regarding your question about `safeApplyShapeAttr`... the `String` shape does not have a `fillOpacity` or `strokeOpacity` attribute defined.

claudep:
OK, thanks, then we are safe as there is already a try/except construct in that loop.

PR #309, opened 2022-01-19
Apply the transformation matrix to the group transform

ChrisBuergi:
Apply the transformation matrix to the group transform by matrix multiplication.

If the "values" is directly assigned to the group transform, then the
the original transformation of the group is lost.

Use the old code if the length of the "values" is not 6. This is somewhat
an undefined state.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -33,6 +33,7 @@ from reportlab.graphics.shapes import (
     _CLOSEPATH, Circle, Drawing, Ellipse, Group, Image, Line, Path, PolyLine,
     Polygon, Rect, SolidShape, String,
 )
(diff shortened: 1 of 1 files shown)


ChrisBuergi:
I do no know how to write a test for this, but here is an example SVG file that renders incorrectly without this patch. The purple rect at the bottom is rendered in the upper left corner without this patch.

[test.svg.txt](https://github.com/deeplook/svglib/files/7903869/test.svg.txt)

claudep:
Thanks a lot for this fix. I added tests and updated Changelog in #311.

ChrisBuergi:
Thanks for the changes. Much appreciated.

PR #310, opened 2022-01-19
Set fill and stroke opacity from color alpha

claudep:
Thanks Chris Buergi for the initial patch.


ChrisBuergi:
Thanks Claude, your solution is more elegant. I like it. Also successfully tested it with my SVG files.

PR #317, opened 2022-02-05
Defines `DEFAULT_FONT_SIZE` variable and fixes None default for `em_base`

### LINKED ISSUE #315: TypeError when using `em` units because `em_base` defaults to `None`
Using `em` units in `width` and `height` of the `svg` tag like this:

```
<svg width="1em" height="1em" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg"></svg>
```

leads to a `TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'`.

<details>
<summary>Full traceback</summary>
<pre>
Traceback (most recent call last):
    drawing = svg2rlg(filename)
  File "svglib/svglib.py", line 1449, in svg2rlg
    drawing = svgRenderer.render(svg_root)
  File "svglib/svglib.py", line 545, in render
    main_group = self.renderSvg(node, outermost=True)
  File "svglib/svglib.py", line 836, in renderSvg
    width, height = self.shape_converter.convert_length_attrs(
  File "svglib/svglib.py", line 957, in convert_length_attrs
    return [
  File "svglib/svglib.py", line 958, in <listcomp>
    convLength(getAttr(attr), attr_name=attr, em_base=em_base, default=default)
  File "svglib/svglib.py", line 321, in convertLength
    return float(text[:-2]) * em_base
TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'
</pre>
</details>

This seems to be caused by `Svg2RlgShapeConverter.convert_length_attrs` defaulting to `em_base=None` and the calling function not setting `em_base`. Would falling back to a general value e.g. `12` like it is done in `Svg2RlgAttributeConverter.convertLength` make sense here?

The idea here is that the `width` and `height` are set relative to the embedding document's `font-size` (which is – of course – not available here).

stefanw:
Fixes #315

### DIFF
--- a/svglib/fonts.py
+++ b/svglib/fonts.py
@@ -18,6 +18,7 @@ STANDARD_FONT_NAMES = (
 DEFAULT_FONT_NAME = "Helvetica"
 DEFAULT_FONT_WEIGHT = 'normal'
 DEFAULT_FONT_STYLE = 'normal'
+DEFAULT_FONT_SIZE = 12
 
 
 class FontMap:
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -54,6 +54,7 @@ from .utils import (
 
 from .fonts import (
     get_global_font_map, DEFAULT_FONT_NAME, DEFAULT_FONT_WEIGHT, DEFAULT_FONT_STYLE,
+    DEFAULT_FONT_SIZE,
 )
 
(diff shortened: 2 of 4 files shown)


claudep (approved):
Looks perfect, thanks!

deeplook (approved):
Great, thanks!

PR #320, opened 2022-02-09
Fix #319 by nudging the first coordinate of some shapes.

### LINKED ISSUE #319: Zero-length line not drawn
Thanks for publishing svglib, it's been really helpful in several of my projects.

## The Problem
I know this sounds odd, but drawing a line of length zero with rounded end caps is an easy way to draw a dot in an SVG file. Unfortunately, the dot disappears when I convert it to PNG with svglib.

The same problem occurs with a polyline of length zero. I haven't tested it, but there might be similar problems with other shapes of size zero.

## Example
Here's a minimal example:

    from pathlib import Path

    svg_text = """\
    <svg baseProfile="full" height="200" version="1.1" width="300"
        xmlns:ev="http://www.w3.org/2001/xml-events"
        xmlns:xlink="http://www.w3.org/1999/xlink">
        <defs />
        <line stroke="#000000" stroke-linecap="round" stroke-width="20"
            x1="150" x2="150" y1="100" y2="100" />
    </svg>
    """

    svg_path = Path('scratch.svg')
    png_path = Path('scratch.png')
    svg_path.write_text(svg_text)
    rlg = svg2rlg(svg_path)
    renderPM.drawToFile(rlg, png_path, fmt="PNG")

When I open the SVG file in a viewer, I see the dot. In the PNG file, I see nothing.

To see the same problem with a polyline, replace the line element with this:

    <polyline stroke="#000000" stroke-linecap="round" stroke-width="20" fill="none"
        points="150,100 150,100" />

## Workaround
If you change the points in the line just enough that they're not equal, then they'll stop getting filtered out. If the change is small enough, the pixels will look the same as the SVG version.

In the example, change the line's x1 value like this:

    <line stroke="#000000" stroke-linecap="round" stroke-width="20"
        x1="150.001" x2="150" y1="100" y2="100" />

## Analysis
You could argue that this is a bug in ReportLab, although I can't find any spec on what should happen to lines of length 0. Since svglib is trying to make image files that match the SVG appearance, I suggest you work around ReportLab's behaviour.

I think the best place for a workaround is in `Svg2RlgShapeConverter.convertLine()` and `Svg2RlgShapeConverter.convertPolyline()`, as well as the conversion functions for any other shapes with the same problem. I suggest checking whether all the points are identical, and then nudging the first one if they are.

I'll try to put together a pull request with this workaround. Let me know if you don't think the fix belongs in svglib, or if you'd prefer a different approach.

donkirkby:
Here's my suggestion on how to stop shapes of size zero from disappearing, even when they have a wide stroke.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -967,8 +967,9 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
         ]
 
     def convertLine(self, node):
-        x1, y1, x2, y2 = self.convert_length_attrs(node, 'x1', 'y1', 'x2', 'y2')
-        return Line(x1, y1, x2, y2)
+        points = self.convert_length_attrs(node, 'x1', 'y1', 'x2', 'y2')
+        nudge_points(points)
+        return Line(*points)
 
     def convertRect(self, node):
         x, y, width, height, rx, ry = self.convert_length_attrs(
@@ -999,6 +1000,7 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
             # Odd number of coordinates or no coordinates, invalid polyline
             return None
 
+        nudge_points(points)
         polyline = PolyLine(points)
         self.applyStyleOnShape(polyline, node)
         has_fill = self.attrConverter.findAttr(node, 'fill') not in ('', 'none')
@@ -1024,6 +1026,7 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
         if len(points) % 2 != 0 or len(points) == 0:
             # Odd number of coordinates or no coordinates, invalid polygon
             return None
+        nudge_points(points)
         shape = Polygon(points)
 
         return shape
@@ -1460,6 +1463,26 @@ def svg2rlg(path, resolve_entities=False, **kwargs):
         os.remove(path)
 
     return drawing
+
+
+def nudge_points(points):
+    """ Nudge first coordinate if all coordinate pairs are identical.
+
+    This works around reportlab's decision to hide shapes of size zero, even
+    when the stroke should be visible.
+    """
+    if not points:
+        return
+    if len(points) < 2:
+        return
+    x0, y0 = points[:2]
+    for i in range(2, len(points)-1, 2):
+        x1, y1 = points[i:i+2]
+        if x1 != x0 or y1 != y0:
+            break
+    else:
+        # All points were identical, so we nudge.
+        points[0] *= 1.0000001
 
 
 def load_svg_file(path, resolve_entities=False):
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -1473,12 +1473,12 @@ def nudge_points(points):
     """
     if not points:
         return
-    if len(points) < 2:
+    if len(points) < 4:
         return
-    x0, y0 = points[:2]
+    x = points[0]
+    y = points[1]
     for i in range(2, len(points)-1, 2):
-        x1, y1 = points[i:i+2]
-        if x1 != x0 or y1 != y0:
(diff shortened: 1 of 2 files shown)


svglib/svglib.py
```diff
@@ -1462,6 +1465,26 @@ def svg2rlg(path, resolve_entities=False, **kwargs):
...
+
+    This works around reportlab's decision to hide shapes of size zero, even
+    when the stroke should be visible.
+    """
+    if not points:
+        return
+    if len(points) < 2:
+        return
+    x0, y0 = points[:2]
+    for i in range(2, len(points)-1, 2):
```
deeplook, line 1479: I wonder if this could be made any faster using built-in functions or sets...?

donkirkby: Sure, I can run some performance tests. What do you think is a reasonable distribution for the number of points in a list? A standard distribution between 10 and 100?

donkirkby: The best performance I found was just removing the unpacking assignments from this code. Using the `all()` function or a set comprehension were both slower.

Here's the performance test I ran:

    from timeit import timeit
    
    ITERATION_COUNT = 900000
    POINT_LISTS = []
    LIST_INDEX = 0
    
    
    def main():
        tests = ['run_test(nudge_points)',
                 'run_test(nudge_points_all)',
                 'run_test(nudge_points_set)']
        for test in tests:
            duration = timeit(test,
                              'setup()',
                              number=ITERATION_COUNT,
                              globals=globals())
            print(duration / ITERATION_COUNT, test)
    
    
    def setup():
        global LIST_INDEX
        POINT_LISTS.clear()
        for i in range(ITERATION_COUNT):
            point_count = 10 + i % 90
            points = [100, 200] * point_count
            if i % 5 != 0:
                # 4/5 of the point lists don't need a nudge.
                points[4] += 1
            POINT_LISTS.append(points)
        LIST_INDEX = 0
    
    
    def run_test(nudge):
        global LIST_INDEX
        points = POINT_LISTS[LIST_INDEX]
        nudge(points)
        LIST_INDEX += 1
    
    
    def nudge_points(points):
        if not points:
            return
        if len(points) < 4:
            return
        x = points[0]
        y = points[1]
        for i in range(2, len(points)-1, 2):
            if x != points[i] or y != points[i+1]:
                break
        else:
            # All points were identical, so we nudge.
            points[0] *= 1.0000001
    
    
    def nudge_points_all(points):
        if not points:
            return
        if len(points) < 4:
            return
        x = points[0]
        y = points[1]
        if all(points[i] == x and points[i+1] == y
               for i in range(0, len(points)-1, 2)):
            # All points were identical, so we nudge.
            points[0] *= 1.0000001
    
    
    def nudge_points_set(points):
        if not points:
            return
        if len(points) < 4:
            return
        points_set = {(points[i], points[i+1]) for i in range(0, len(points)-1, 2)}
        if len(points_set) == 1:
            # All points were identical, so we nudge.
            points[0] *= 1.0000001
    
    
    main()

The results were:

    1.6128685722247206e-06 run_test(nudge_points)
    2.4019194488856658e-06 run_test(nudge_points_all)
    7.203874138892085e-06 run_test(nudge_points_set)

Let me know if you'd like me to try any other approaches.

For now, I'll push the new version without the unpacking assignments.

deeplook: Fine then, thanks!

donkirkby:
Thanks, @deeplook, I don't plan to make any other changes.
I sent a message to the ReportLab mailing list, but I haven't seen any response. If they do agree to change their behaviour, it would be easy to reverse this change.

deeplook:
Thanks, again!

PR #325, opened 2022-03-16
Guard against empty style tags

stefanw:
An SVG with emtpy style tag like `<style></style>` breaks parsing because `node.text` is `None` and  `CSSMatcher`  doesn't like a `None` argument.

This PR fixes this by passing in an empty string if `node.text` is Falsey. Test included.


claudep:
Excellent, thanks!

PR #326, opened 2022-03-22
Allow multiple style nodes

dbasedow:
I am dealing with SVGs generated by a tool that puts a style tag for each class into defs. At the moment the result of using svg2rlg is that only the last style element is used.

This PR changes the behavior of renderStyle to append additional rules to the CSSMatcher that was created during initialization of SvgRenderer.


claudep:
I don't know why CI tests don't run. Maybe you can try to squash and force push to see if it triggers CI tests?

Issue #343, opened 2022-05-19
Rendering issue with <tspan> and <style>


jspobst:
A plotly figure I made isn't rendering correctly. The part that's failing involves a `<tspan>`, so I went and grabbed the simplest tspan example I could find from [here](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/tspan) to test further.

```
<svg viewBox="0 0 240 40" xmlns="http://www.w3.org/2000/svg">
  <style>
    text  { font: italic 12px serif; }
    tspan { font: bold 10px sans-serif; fill: red; }
  </style>

  <text x="10" y="30" class="small">
    You are
    <tspan>not</tspan>
    a banana!
  </text>
</svg>
```

That _should_ produce 
<img width="274" alt="image" src="https://user-images.githubusercontent.com/1304503/169291457-e9657f39-baf4-424b-ad91-6831f073e680.png">

But when putting it through 
```
drawing = svglib.svg2rlg(io.StringIO(textwrap.dedent(svg_string)))
renderPDF.draw(drawing, c, 50, 50)
```

I get this in the produced pdf 
<img width="274" alt="image" src="https://user-images.githubusercontent.com/1304503/169292078-6705d85b-5c23-449f-a78d-40fb59909afc.png">

svglib: 1.3.0
reportlab: 3.6.9
python: 3.97
MacOS: 12.3.1

Thank you.

jspobst:
Given that no text after `not` shows, it almost seems as though the `</tspan>` is somehow closing the `<text>`.

claudep:
In #344, I tried to solve the issue of the trailing part of text nodes. The rendering of the sample above is not yet optimal, but at least it seems all text appear.

jspobst:
Hi @claudep, I see that this isn't optimal just yet, but I wanted to say thanks for the work you've already put into it so quickly. It really is appreciated!

PR #344, opened 2022-05-19
Refs #343 - Don't forget to render tail of text nodes

claudep:


claudep:
The patch may need some more work.

claudep:
@jspobst, if you want to test this PR, you are welcome!

jspobst:
I just checked it out, and tested the example code I put in the issue. I can confirm that all of the text shows, but the formatting (ie, italics and bold) still doesn't work. The result: 
image

If this PR was meant only to address some of the text missing before, I'd say it looks good (but doesn't fully close the corresponding issue). Admittedly, I might not understand all of what goes into 'testing this PR', so if that's more than simply generating one pdf, further testing by someone who understands best practices might be needed.

claudep:
I thought you might have other real SVGs that you might test.

The style issue is a completely different thing. svglib doesn't yet support the shorthand `font`, but only the more specific `font-*` properties.

PR #355, opened 2022-08-12
"ex" and "ch" must be assumed as em/2

thisismiller:
My interest in this is specifically support for `ch`, as it's the only way to get SVGs to be able to get lines aligned with text well.

W3 offers guidance on what to do when you can't inspect the text to determine the x-height (`ex`) or the advance measure of a `0` glyph (`ch`).
* In the cases where it is impossible or impractical to determine the x-height, a value of 0.5em must be assumed.
* In the cases where it is impossible or impractical to determine the measure of the “0” glyph, it must be assumed to be 0.5em wide by 1em tall.

There's precedent in both [CairoSVG](https://github.com/Kozea/CairoSVG/blob/49c53b3a1d59d8ba1236095505e6ffb70f9ddfcb/cairosvg/helpers.py#L379) and [ImageMagick](https://github.com/ImageMagick/ImageMagick/blob/b42d5cbea9bb289130094d6299ff4897b75ab37b/coders/svg.c#L725) which have already adopted `em / 2.0` as their way to interpret `ex`, so with this change, I'm only suggesting that svglib do the same.

Neither currently support `ch`, but it's good to follow W3's advice.

---

Here's the motivating case of an SVG I made:

ophistory_all

This probably renders fine, as you're looking at it from a webbrowser, which understands how to render `ch` correctly.

Currently, processing this SVG into a png with svglib+reportlab gives an error like:

```
Traceback (most recent call last):
  File "/home/miller/ws/ophistory/./ophistory.py", line 354, in <module>
    main(sys.argv)
  File "/home/miller/ws/ophistory/./ophistory.py", line 348, in main
    drawing = svg2rlg(f)
  File "/home/miller/.local/lib/python3.10/site-packages/svglib/svglib.py", line 1404, in svg2rlg
    drawing = svgRenderer.render(svg_root)
  File "/home/miller/.local/lib/python3.10/site-packages/svglib/svglib.py", line 502, in render
    view_box = self.get_box(node, default_box=True)
  File "/home/miller/.local/lib/python3.10/site-packages/svglib/svglib.py", line 768, in get_box
    width, height = map(self.attrConverter.convertLength, (width, height))
  File "/home/miller/.local/lib/python3.10/site-packages/svglib/svglib.py", line 334, in convertLength
    length = toLength(text)  # this does the default measurements such as mm and cm
  File "/home/miller/.local/lib/python3.10/site-packages/reportlab/lib/units.py", line 30, in toLength
    raise ValueError("Can't convert '%s' to length" % s)
ValueError: Can't convert '197ch' to length
```

With the change I have proposed, the diagram renders to a .png via svglib like:

ophistory_all_svglib

And so even if assuming `0.5em` is slightly inaccurate, is such an incredible improvement that I think it justifies using `ex` as an approximation for `ch`.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -325,10 +325,14 @@ class Svg2RlgAttributeConverter(AttributeConverter):
             return float(text[:-2]) * em_base
         elif text.endswith("px"):
             return float(text[:-2])
-
-        if "ex" in text:
-            logger.warning("Ignoring unit ex")
-            text = text.replace("ex", '')
+        elif text.endswith("ex"):
+            # The x-height of the text must be assumed to be 0.5em tall when the
+            # text cannot be measured.
+            return float(text[:-2]) * em_base / 2
+        elif text.endswith("ch"):
+            # The advance measure of the "0" glyph must be assumed to be 0.5em
+            # wide when the text cannot be measured.
+            return float(text[:-2]) * em_base / 2
 
         text = text.strip()
         length = toLength(text)  # this does the default measurements such as mm and cm
--- a/tests/samples/others/em_unit.svg
+++ b/tests/samples/others/em_unit.svg
@@ -1,1 +1,0 @@
-<svg width="1em" height="1em" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg"></svg>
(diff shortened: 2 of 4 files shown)


claudep (approved):
Thanks a lot for the patch and the well-described report!

PR #361, opened 2022-11-18
Added support for "jpg" base64 encoded images

FelixFrog:
Previously this would create problems for certain svgs, making image not render

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -685,7 +685,7 @@ class SvgRenderer:
             return None
 
         # First handle any raster embedded image data
-        match = re.match(r"^data:image/(jpeg|jpg|png);base64", xlink_href)
+        match = re.match(r"^data:image/(jpe?g|png);base64", xlink_href)
         if match:
             img_format = match.groups()[0]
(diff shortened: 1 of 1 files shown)


deeplook:
Does it make sense to add "jpeg" (with an "e") as well?

svglib/svglib.py
```diff
@@ -685,7 +685,7 @@ def xlink_href_target(self, node, group=None):
             return None
 
         # First handle any raster embedded image data
-        match = re.match(r"^data:image/(jpeg|png);base64", xlink_href)
+        match = re.match(r"^data:image/(jpeg|jpg|png);base64", xlink_href)
```
claudep, line 688: As we are in a regex already, I would simply put a `?` after the `e`.

FelixFrog: Yes, you're right

FelixFrog:
> Does it make sense to add "jpeg" (with an "e") as well?

I think the code already works for ```href="data:image/jpeg;base64,..."``` images. It just didn't process the ones where jpg is spelled without and "e", resulting in incomplete renders.

claudep:
Thanks!

PR #365, opened 2022-12-07
Fix conversion from shorthand quadratic to cubic bézier

### LINKED ISSUE #364: Implementation error in rendering of paths with shorthand command for quadratic Bézier curves
I noticed weird SVG curves coming out of the internal conversion from quadratic Bézier curves to cubic ones.

[Here's an example of what the mistake can looks like in practice.](https://svg-path-visualizer.netlify.app/#M%0A257.8%2C17.41Q%0A257.8.76%2C269.86.77t%0A12%2C16.64q%0A0%2C16.72-12%2C16.73T%0A257.8%2C17.41Z%0AM%0A257.8%2C17.41C%0A257.8%2C6.31%2C261.82%2C0.763333%2C269.86%2C0.77C%0A275.22%2C0.774444%2C279.22%2C6.321111%2C281.86%2C17.41C%0A281.86%2C28.556667%2C277.86%2C34.133333%2C269.86%2C34.14C%0A264.526667%2C34.144444%2C260.506667%2C28.567778%2C257.8%2C17.41Z)
- The first "M ... Z" uses quadratic curves and results in a nice oval "O".
- The second "M ... Z" uses cubic curves generated from `svglib` where the control points of the previous `t`/`T` commands are off.

I tracked this down to [this code in `convertPath`](https://github.com/deeplook/svglib/blob/9a43d4f71a483f844b905d12682d35b18d9099a7/svglib/svglib.py#L1168-L1204). The [definition of the `t`/`T` commands](https://www.w3.org/TR/SVG/paths.html#PathDataQuadraticBezierCommands) is:
> Draws a quadratic Bézier curve from the current point to (x,y). The control point is assumed to be the reflection of the control point on the previous command relative to the current point.
> (If there is no previous command or if the previous command was not a Q, q, T or t, assume the control point is coincident with the current point.)

So where does the control point from the previous command come from? `svglib` takes the previously stored last two points and assumes the first one to be the control point of the previous quadratic Bézier curve. However, the point is actually the second control point of a *cubic* Bézier curve (because a previous `Q`/`q` command has been converted to cubic).

The solution would be to store the control point of the last quadratic Bézier curve in a variable and use that when the next `t`/`T` command comes along. I will prepare a PR for this.

stefanw:
Closes #364

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -1094,6 +1094,7 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
         unclosed_subpath_pointers = []
         subpath_start = []
         lastop = ''
+        last_quadratic_cp = None
 
         for i in range(0, len(normPath), 2):
             op, nums = normPath[i:i+2]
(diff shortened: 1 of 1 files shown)


deeplook:
Thanks for this patch! This looks good to me, but I'm notsure why the CI fails.

stefanw:
I rebased the Actions changes and test pass now (apart from DeepSource which seems not relevant).

Issue #366, opened 2022-12-09
Transform on clipPath does not work


blayzen-w:
When clipPaths contain a path with a transform, the path is ignored and discarded with the error message `Unsupported shape type Group for clipping`.

The svg I am testing with is the following:
```
<svg namespace="http://www.w3.org/XML/1998/namespace" version="1.1" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <defs>
        <clipPath id="clip-1">
            <path transform="scale(0.5, 0.5)" d="M 50, 50 m -50, 0 a 50,50 0 1,0 100,0 a 50,50 0 1,0 -100,0"/>
        </clipPath>
    </defs>
    <g>
        <g clip-path="url(#clip-1)">
            <path d="M25,25h-50v-50h50V25z"/>
        </g>
    </g>
</svg>
```

I don't believe this was happening with earlier versions so I decided to do some digging and narrowed the problem down.

The error that is displayed starts here, when the detected shape from the node is neither a Path or a Rect despite the svg using a path element.
https://github.com/deeplook/svglib/blob/9a43d4f71a483f844b905d12682d35b18d9099a7/svglib/svglib.py#L655-L656

This is caused from the convertShape function where, if there is a transform, then it will add the shape to a group and apply the transform to the group instead of the shape.
https://github.com/deeplook/svglib/blob/9a43d4f71a483f844b905d12682d35b18d9099a7/svglib/svglib.py#L895-L916

I did a quick test and applied the transform directly on the path element and everything seemed to work fine, however I'm not sure if there are any unintended side effects from doing it this way. Is this an intended feature or could I work on a patch to add an optional argument to apply the transformation directly for this use case?

claudep:
I don't remember if the fact to apply transforms only on shapes was a conveniency, or if it was required. Feel free to experiment and suggest changes if they are successful.

PR #367, opened 2022-12-09
Issue #366 - Add support for transforms on clipPath elements

blayzen-w:
See Issue #366 for details.

Let me know if you would like anything changed.

I want to validate this on a few more larger files before I move it out of draft mode.

### DIFF
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -337,7 +337,18 @@ class TestW3CSVG:
                     with open(join(TEST_ROOT, "samples", archive_path), "wb") as f:
                         f.write(data)
                 print(f"unpacking {archive_path}")
-                tar_data = gzip.open(join(TEST_ROOT, "samples", archive_path), "rb").read()
+
+                # Open normally or use gzip if compressed
+                def opener(filename):
+                    f = open(filename, 'rb')
+                    if f.read(2) == '\x1f\x8b':
+                        f.seek(0)
+                        return gzip.GzipFile(fileobj=f)
+                    else:
+                        f.seek(0)
+                        return f
+
+                tar_data = opener(join(TEST_ROOT, "samples", archive_path)).read()
                 with open(join(TEST_ROOT, "samples", tar_path), "wb") as f:
                     f.write(tar_data)
             print(f"extracting into {self.folder_path}")
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -615,15 +615,17 @@ class SvgRenderer:
         def get_shape_from_node(node):
             for child in node.iter_children():
                 if node_name(child) == 'path':
-                    group = self.shape_converter.convertShape('path', child)
-                    return group.contents[-1]
+                    group = self.shape_converter.convertShape('path', child, transform_shape=True)
+                    return group.contents[-1], group.transform
                 elif node_name(child) == 'use':
                     grp = self.renderUse(child)
-                    return get_shape_from_group(grp)
+                    return get_shape_from_group(grp), None
                 elif node_name(child) == 'rect':
-                    return self.shape_converter.convertRect(child)
+                    return self.shape_converter.convertRect(child), None
                 else:
                     return get_shape_from_node(child)
+
+            return None, None
 
         clip_path = node.getAttribute('clip-path')
         if not clip_path:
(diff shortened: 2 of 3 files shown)


blayzen-w:
The only issue that I have found so far is that scale transforms don't seem to be sandboxed correctly (translate works correctly). At this point I'm not sure if this is an issue with reportlab or svglib. I'll take a look at this some more on Monday but I'll put my findings below.

Take the following svg:
```
<svg namespace="http://www.w3.org/XML/1998/namespace" 
     xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" 
     viewBox="0 0 100 100">
     
    <defs>
        <clipPath id="clip-1">
            <path transform="scale(2,1)" d="M 50, 50 m -50, 0 a 50,50 0 1,0 100,0 a 50,50 0 1,0 -100,0"/>
        </clipPath>
    </defs>
    <g transform="scale(1,3)">
        <g clip-path="url(#clip-1)">
            <path d="M50,50H0V0h50V50z" fill="red"/>
        </g>
    </g>
</svg>
```

The square should be 50x150, however its being generated at 100x150. It almost seems that the clip path scale it being applied to the square that its clipping.

tests/test_samples.py
```diff
@@ -337,7 +337,18 @@ def setup_method(self):
...
+                def opener(filename):
+                    f = open(filename, 'rb')
+                    if f.read(2) == '\x1f\x8b':
+                        f.seek(0)
+                        return gzip.GzipFile(fileobj=f)
+                    else:
+                        f.seek(0)
+                        return f
+
+                tar_data = opener(join(TEST_ROOT, "samples", archive_path)).read()
```
claudep, line 351: May I ask you to put this part in a separate PR, please?

claudep: I may work on this now, in fact, wait a bit.

claudep: See #368. You can drop this part of the patch.

blayzen-w: Yeah, no problem. I just wanted to make sure the test cases were passing.

blayzen-w:
It looks like that issue that I'm running into is a bug with reportlab.

From what I can tell, the transformation is being applied when rendering the clipPath element, however when it's finished the state is never unapplied. So the transformation is then applied to the next element that is rendered. This can be seen in the drawNode function where it doesn't save and restore the canvas state if it's a path with clipPath set (`if not (isinstance(node, Path) and node.isClipPath)`).

```
    def drawNode(self, node):
        """This is the recursive method called for each node
        in the tree"""
        #print "pdf:drawNode", self
        #if node.__class__ is Wedge: stop
        if not (isinstance(node, Path) and node.isClipPath):
            self._canvas.saveState()

        #apply state changes
        deltas = getStateDelta(node)
        self._tracker.push(deltas)
        self.applyStateChanges(deltas, {})

        #draw the object, or recurse
        self.drawNodeDispatcher(node)

        self._tracker.pop()
        if not (isinstance(node, Path) and node.isClipPath):
            self._canvas.restoreState()
```

When the check is removed and the canvas state is saved and restored, the element is sized correctly but the clip path is not applied. They don't seem to host their code publicly so hopefully they are receptive to changes/bug reports from the public. Otherwise maybe we might be able to monkey patch it but I don't care for those too much.

blayzen-w:
I haven't heard anything from reportlab so I did a bit of digging in the PDF ISO 32000-1 spec and found this interesting note.

`There is no way to enlarge the current clipping path or to set a new clipping path without reference to the current one. However since the clipping path is part of the graphics state, its effect can be localized to specific objects by enclosing the modification of the clipping path and the painting of those objects between a pair of q and Q operators.` - page 138.

So if I'm understanding this correctly, trying to apply a transform on the clipping path may be a dead end, since the clipping path transform needs to not depend on the current graphical state since its referenced. I have almost no experience working with pdfs so I could be misunderstanding things, and I don't have access to ISO 32000-2 to see if anything has changed since 2008.

I was hoping this would be an easy fix since Illustrator seems to import svgs just fine, but I guess not. I'm not sure when I will have time to continue working on this. If you want I can close the draft and open a new one once I find a better solution.

claudep:
Thanks a lot for your explorations. Feel free to close or to let open depending on your future plans.

PR #378, opened 2023-06-02
Add pyproject.toml

claudep:
setup.py is obsolete nowadays and replaced by pyproject.toml (PEP 518).


deeplook:
I'm fine with moving to pyproject.toml, although I'm quite sure that support for setup.py will not vanish really soon. But this PR removes instructions for installation without pip from a locally downloaded package, though, if I'm not mistaken. Or am I missing something?

claudep:
Instead of removing the section, I should have moved the downloaded package installation from `setup.py` to `pip`.

Anyway, I think there is a unrelated reportlab installation issue in the test suite, so putting this as WIP until it's fixed.

rettinghaus:
When introducing pyproject you should follow [PEP 621](https://peps.python.org/pep-0621/) adding all metadata and add all dependencies, removing [requirements.txt](https://github.com/deeplook/svglib/blob/main/requirements.txt).

claudep:
It's not yet very clear to me what's the exact equivalent of `pip install -r requirements.txt` when you moved your deps to `pyproject.toml`.

deeplook:
> It's not yet very clear to me what's the exact equivalent of `pip install -r requirements.txt` when you moved your deps to `pyproject.toml`.

It depends on the package manager you want to use, today most likely `poetry` or `uv`. Both have, of course, slightly different ways of organizing dependencies in `pyproject.toml`. But it would be nice to support both in the same file.

claudep:
Feel free to close and create your own PR to finish this.

PR #381, opened 2023-06-24
use fillMode directly instead of _fillRule and reportlab monkeypatch

MrBitBucket:
Hi Dinu & Claude, here is my patch to eliminate the reportlab monkeypatch

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,8 @@ docs/_build
 env/
 .workon
 dist/
-tests/samples/misc/*-svglib.pdf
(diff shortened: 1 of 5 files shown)


deeplook:
Applied in [#b2fdd58](https://github.com/deeplook/svglib/commit/b2fdd581a590ec128e58dec584a7112c0c592451) — thanks @MrBitBucket! This was a superset of #440 which was merged separately; the remaining pieces (monkeypatch removal, SVG test samples, improved debug logging, and .gitignore cleanup) are now in main.

Issue #382, opened 2023-06-24
CI failure because of missing cairo


claudep:
Tests on GitHub CI fail currently because of missing cairo dependencies (and hence png output is not available).

claudep:
We may set `reportlab[pycairo]` in requirements.txt, but should we also add the ability to optionally install it, like reportlab, by using `extra_require`?

deeplook:
We could set a hard dependecy on `pycairo` only for `reportlab>=4.0.0` like mentioned in [this example](https://stackoverflow.com/questions/73371336/setup-pys-install-requires-how-do-i-specify-python-version-range-for-a-specif).

claudep:
My question was about keeping pycairo as optional even with recent reportlab. As cairo installation is not trivial on some platforms, maybe people who just want to use svglib to produce pdf (and no png), cairo installation may not be a hard dependency.

deeplook:
Yes, I agree it should be optional with a hint to get it installed for reportlab>=4 and the trouble to expect. Still, it would be nice to have it work inside github actions, though.

MrBitBucket:
The problem with actually using rlPyCairo (ie pycairo) is that pycairo does not build installable wheels for most platforms. That forces all the linux / unixy platforms to compile their own. 

Part of the reason for splitting the bitmap stuff out of reportlab was to reduce the skills required to maintain it going forward. I can just about hack the github cibuildwheel stuff to obtain windows/manylinux/osx wheels which we did up to 4.0, but my continued participation is not guaranteed.

If testing is all that's needed then a good start would be 

https://github.com/pygobject/pycairo/blob/main/.github/workflows/test.yml

which actually needs to build pycairo prior to testing it. Presumably svglib with reportlab+rlPyCairo would need to build/install libcairo(-dev), pycairo, rlPyCairo & reportlab. 

I got very fed up with yml doing wheel building for reportlab, I have no doubt the same will be true for testing.

deeplook:
@MrBitBucket I guess we could limit the tests run by GitHub actions only to those platforms known to work fine with pycairo. I remember I had some issues with it in a different context, but I forgot the details. I've tried to find out which warnings regarding pycairo I'd see in the installation descriptions in reportlab 4.x, and I could not find any in the archive I get from https://pypi.org/project/reportlab/#files (there is only a sentence naming pycairo as a dependency for bitmap images with text).

[...]

PR #385, opened 2023-08-17
fix: using dedicated logger instead of root logger

### LINKED ISSUE #384: Using root log instead of logger
https://github.com/deeplook/svglib/blob/774224413af53c16f23df76b4edc0c944968bd68/svglib/svglib.py#L656

When this error is triggered, the log is published to the root logger, instead of to the module logger.

kiinami:
Closes #384


claudep:
Sorry, #382 struck :cry:

claudep:
I'll merge it anyway.

claudep:
Thanks! :1st_place_medal:

Issue #391, opened 2023-11-20
Add more example and pip install commands to README.md


dinhanhx:
For [example section](https://github.com/deeplook/svglib#examples), the following code can be added to demonstrate how to convert a svg file to a Pillow Image object

Example svg file: https://upload.wikimedia.org/wikipedia/commons/4/4f/SVG_Logo.svg

```python
from svglib.svglib import svg2rlg
from reportlab.graphics import renderPM
from PIL import Image
import io

drawing = svg2rlg("SVG_Logo.svg")
place_holder = io.BytesIO()
renderPM.drawToFile(drawing, place_holder, fmt="PNG")

image = Image.open(place_holder)
image.show()
```

For [pip install section](https://github.com/deeplook/svglib#1-using-pip), I believe it would be better if this was added

```bash
pip install svglib rlPyCairo reportlab
```

So other people can have quick installation without fixing missing dependencies.

claudep:
Thanks for this suggestion, would you like to create a pull request with this snippet?

dinhanhx:
I believe it would be better if you do that because you might want to edit the suggested examples.

deeplook:
We can easily adjust it after you add a contribution to this package and your own list. ;-)

deeplook:
Closing as stale — no PR ever materialized for this suggestion. Happy to reconsider if someone wants to submit the README addition.

PR #392, opened 2023-12-03
Add more example and pip install commands to README.md

dinhanhx:
#391

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -109,6 +109,22 @@ Note that the second parameter of ``drawToFile`` can be any
 `Python file object`_, like a ``BytesIO`` buffer if you don't want the result
 to be written on disk for example.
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -109,6 +109,22 @@ Note that the second parameter of ``drawToFile`` can be any
 `Python file object`_, like a ``BytesIO`` buffer if you don't want the result
 to be written on disk for example.
 
+You can convert a svg file to pillow image object as following snippet:
+
+.. code:: python
+
+    from svglib.svglib import svg2rlg
```
deeplook, line 116: Nitpicking, but this order would be a more recommended one: stdlib, external, and local packages, not split with empty lines here for brevity:

```python
import io
from PIL import Image
```

Issue #393, opened 2023-12-15
svglib doesn't remove temporary files properly?


wgundamj44:
When using svglib on svgs that have images, we found the disk size keeps growing up. The culprit is the temporary files created by this method `xlink_href_target`.

This line https://github.com/deeplook/svglib/blob/main/svglib/svglib.py#L701 mentioned the temporary files needs to be removed later, but I didn't find the place where they got removed. 

Did I miss something or the removal is really missing?

claudep:
It's totally possible the file is never removed. Any implementation suggestion is welcome.

wgundamj44:
I think instead of creating a temporary file here https://github.com/deeplook/svglib/blob/main/svglib/svglib.py#L695, maybe it can create a `PIL.Image` instead. From what I read, `reportlab.graphcis.renderPDF.drawImage`  can accept `PIL.Image` in new versions.

claudep:
Any pull request is welcome!

claudep:
Fixed by 484d3660a34cfedb24431aed8ecbb61a30afb193

PR #394, opened 2023-12-15
make xlink_href_target method create PIL image instead of temporary file

wgundamj44:
Described in https://github.com/deeplook/svglib/issues/393

The current implementation creates a temporary file and pass its path all the way down to reportlab for render.
It's hard to find a place to unlink the temporary file because it might be outside of svglib's reach when reportlab is done with the file.

This PR tries to pass PIL Image to reportlab, since reportlab seems to support both file path and Image.

### DIFF
--- a/svglib/svglib.py
+++ b/svglib/svglib.py
@@ -21,10 +21,11 @@ import logging
 import os
 import pathlib
 import re
-import tempfile
 import shlex
 import shutil
+from io import BytesIO
 from collections import defaultdict, namedtuple
+from PIL import Image as PILImage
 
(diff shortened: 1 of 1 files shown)


deeplook:
Thanks! Could you also please add a test case?

claudep:
I think that part of the code is already covered by `test_samples.TestOtherFiles.test_png_in_svg`. @deeplook, did you have another test in mind?

deeplook:
@claudep True, that should do it! Fine then with me.

claudep (approved):
Thanks, I checked the test suite passes locally on Linux.

Issue #397, opened 2024-01-11
svglib ignores <switch> nodes


ilkoch008:
Is it possible to add support for at least default cases (without ```requiredFeatures```)?

My case:
I use rst2pdf to compile document containing svg images.
Those svg images are exported from draw.io.
draw.io uses ```<switch>``` when exporting to svg for all text blocks.

Example:
```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="121" height="61" viewBox="-0.5 -0.5 121 61">
  <rect width="120" height="60" rx="9" ry="9" fill="#FFF" stroke="#000" pointer-events="all"/>
  <switch transform="translate(-.5 -.5)">
    <foreignObject pointer-events="none" width="100%" height="100%" requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility" style="overflow:visible;text-align:left">
      <div xmlns="http://www.w3.org/1999/xhtml" style="display:flex;align-items:unsafe center;justify-content:unsafe center;width:118px;height:1px;padding-top:30px;margin-left:1px">
        <div data-drawio-colors="color: rgb(0, 0, 0);" style="box-sizing:border-box;font-size:0;text-align:center">
          <div style="display:inline-block;font-size:12px;font-family:Helvetica;color:#000;line-height:1.2;pointer-events:all;white-space:normal;overflow-wrap:normal">
            some text
          </div>
        </div>
      </div>
    </foreignObject>
    <text x="60" y="34" font-family="Helvetica" font-size="12" text-anchor="middle">some text</text>
  </switch>
</svg>
```

As a result, all text in the images is lost after assembly.

claudep:
Possible, certainly, but someone will need to add this support.

ilkoch008:
Found out that just default case processing is not enough for me, but maybe this would help someone.

deeplook:
Just to add some context: https://www.w3.org/TR/SVG/struct.html#SwitchElement

deeplook:
Should be implemented in https://github.com/deeplook/svglib/pull/441.

PR #398, opened 2024-01-12
Add simple switch process

ilkoch008:
#397 
This pull request adds simple support of ```switch``` nodes (choose first child without ```requiredFeatures```, i.e. default case).


claudep:
Thanks for the patch! It would still need a test, and probably a new line under `Known limitations` in the README, telling about limited switch support.

deeplook:
Implemented now in https://github.com/deeplook/svglib/pull/441.

PR #426, opened 2026-01-09
Update svg2rlg to accept IO as path type

ThexXTURBOXx:
On my current project, this is the only warning I get from PyCharm.
This seems to be officially supported since #173/#174

### DIFF
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -30,7 +30,7 @@ import shlex
 import shutil
 from collections import defaultdict, namedtuple
 from io import BytesIO
-from typing import Any, BinaryIO, Dict, List, Optional, Tuple, Union
+from typing import Any, BinaryIO, Dict, List, Optional, TextIO, Tuple, Union
 
 from PIL import Image as PILImage
 from reportlab.graphics.shapes import (
@@ -1782,7 +1782,7 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
 
 
 def svg2rlg(
-    path: Union[str, os.PathLike[str], BinaryIO],
+    path: Union[str, os.PathLike[str], BinaryIO, TextIO],
     resolve_entities: bool = False,
     **kwargs: Any,
 ) -> Optional[Drawing]:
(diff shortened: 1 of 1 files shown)


src/svglib/svglib.py
```diff
@@ -30,7 +30,7 @@
 import shutil
 from collections import defaultdict, namedtuple
 from io import BytesIO
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import IO, Any, Dict, List, Optional, Tuple, Union
```
deeplook: Please change this to `BinaryIO`, since `path` in the function `svg2rlg` can also be a binary SVGZ file, and even for SVG it would still be correct. Also, please don't break the alphabetical order, or some tools might complain (I hope).

Also, please consider writing a test (not sure there are any for this, already).

ThexXTURBOXx: Interesting - the order was changed by the pre-commit hook: https://github.com/deeplook/svglib/pull/426/changes/4d7728a46b0ebf477b948846136379b3634665d5 (my commit had the correct order: https://github.com/deeplook/svglib/pull/426/changes/78a593babc66b94bd0c8528670778f3ff2b2de7e).
But apart from that, I fixed the alphabetical order and changed it to `BinaryIO`. Thanks for the feedback!
Even though, why shouldn't `StringIO` be supported? According to the already existing tests, this seems to also be supported (see e.g., [here](https://github.com/deeplook/svglib/blob/main/tests/test_basic.py#L92-L94) or [a few years back here](https://github.com/deeplook/svglib/pull/174/changes#diff-29093b485f46bc37dda5e0854cfc54e289f24e206207afb81dff1c527cb001a7R656)).

deeplook: It's `typing.TextIO`, not `typing.StringIO`, but yes, let's add that, too.

ThexXTURBOXx: Done!

deeplook:
Merged, thanks! There are some unrelated build errors, partly related to access limitations from Wikimedia, that I'll look into at some point.

PR #429, opened 2026-01-28
Fixes #428 - Improved Fontconfig loading for font variants

### LINKED ISSUE #428: Better bold/italic font detection by fontconfig
Currently, when searching fonts through font-config `fc-search`, the returned path does not search for specific bold or italic versions of fonts. However, the `fc-search` syntax allows for specifying font-variants. I'll suggest a patch.

claudep:


claudep:
I did not include tests because the fontconfig stuff is a bit tricky to test. If you think we should still add some test, please give some ideas.

deeplook:
Makes sense to me, although the exact impact is hard to estimate and likely test as you say. So, LGTM.

claudep:
Thanks!

Issue #449, opened 2026-04-28
Add support for SVG 2 viewport and root-relative units: rem, vw, vh, vmin, vmax, q


deeplook:
## Summary

svglib does not implement the CSS/SVG 2 length units introduced after SVG 1.1.
These units are increasingly common in modern SVG content.

## Missing units

| Unit | Meaning | Resolution strategy |
|------|---------|-------------------|
| `rem` | Root element font-size | Capture root `font-size` during render and use as `rem` base |
| `vw` | 1% of viewport width | Resolve against `main_box.width` (already tracked) |
| `vh` | 1% of viewport height | Resolve against `main_box.height` (already tracked) |
| `vmin` | Smaller of `vw` and `vh` | `min(vw, vh)` |
| `vmax` | Larger of `vw` and `vh` | `max(vw, vh)` |
| `q` | Quarter-millimetre (0.25 mm) | Straightforward physical conversion |

`ch` is already approximated as 0.5em.

## Notes

- `vw`/`vh` should treat the outermost SVG canvas as the viewport (the PDF
  render target has no browser viewport).
- `rem` requires storing the computed root font-size on `Svg2RlgAttributeConverter`
  before the tree walk begins.
- The `convertLength` method in `Svg2RlgAttributeConverter` already has the
  right structure to add these as new `elif` branches.
- Target: 2.1 (after 2.0 coordinate-system change stabilises).

Issue #448, opened 2026-04-28
preserveAspectRatio not implemented — non-uniform scaling distorts SVGs with mismatched viewBox/canvas aspect ratios


deeplook:
## Summary

svglib applies independent x and y scales when mapping a `viewBox` to the canvas dimensions, but does not implement the `preserveAspectRatio` attribute. This causes non-uniform stretching whenever the viewBox aspect ratio differs from the width/height aspect ratio.

## Affected example

`tests/samples/misc/arcs01.svg`:
```xml
<svg width="12cm" height="5.25cm" viewBox="0 0 1200 400" ...>
```

- viewBox aspect ratio: 1200 / 400 = **3.0**
- Canvas aspect ratio: 12cm / 5.25cm ≈ **2.286**

svglib computes:
- x_scale = (12 cm → pt) / 1200 ≈ **0.2835**
- y_scale = (5.25 cm → pt) / 400 ≈ **0.3720**

The non-equal scales distort the output: circles become ellipses, squares become rectangles, etc.

## Expected behaviour (SVG spec §10.8)

When `preserveAspectRatio` is absent or `"xMidYMid meet"` (the default), a **uniform** scale should be used:

```
uniform_scale = min(canvas_width_pt / viewBox.width,
                    canvas_height_pt / viewBox.height)
```

The scaled content is then centred within the canvas. Non-uniform scaling (`none`) should only be applied when `preserveAspectRatio="none"` is explicitly set.

## Impact

Any SVG whose viewBox proportions do not exactly match its width/height proportions will be silently distorted. This is a very common pattern in real-world SVGs.

## Suggested fix

In `renderSvg` (around the `group.scale(x_scale, y_scale)` call), read the `preserveAspectRatio` attribute and:

1. Parse the `<align> <meetOrSlice>` tokens.
2. For `meet` (default): `scale = min(x_scale, y_scale)`; translate to align per the align token.
3. For `slice`: `scale = max(x_scale, y_scale)`; clip to canvas bounds.
4. For `none`: keep independent x/y scales (current behaviour).

## Note

This bug predates the coordinate-system refactor in PR #… and is independent of the bare-number / `px` unit fix.

Issue #433, opened 2026-04-15
Bold font weight and font size are changing, when using `svg2rlg`


mosc9575:
I have a valid svg generated with bokeh and I try to convert this to an PDF or EPS using `svglib` and `reportlab`.

I reduced the example to a bold title.

<img width="150" height="40" alt="Image" src="https://github.com/user-attachments/assets/6ec06453-798c-403f-8b8e-2405f810906c" />

When converting this text, the expected result is a bold font in the same size, but I got this.

[title_example.pdf](https://github.com/user-attachments/files/26753850/title_example.pdf)

Minimal example:

```python
from io import StringIO
from svglib import fonts, svglib
from reportlab.graphics.renderPDF import drawToFile

# register font to make sure no other problems appear
fonts.register_font_family(
    "helvetica",
    normal=r"C:\Windows\Fonts\Helvetica\Helvetica.tff",
    bold=r"C:\Windows\Fonts\Helvetica\Helvetica-Bold.tff",
    italic=r"C:\Windows\Fonts\Helvetica\Helvetica-Oblique.tff",
    bolditalic=r"C:\Windows\Fonts\Helvetica\Helvetica-Boldoblique.tff",
)

s = """<svg version="1.1" xmlns="http://www.w3.org/2000/svg" width="150" height="40">
    <text fill="#444444" stroke="none" font-family="helvetica" font-size="13px" font-style="normal"
        font-weight="bold" text-decoration="normal" x="0" y="12" text-anchor="start"
        dominant-baseline="alphabetic" transform="matrix(1, 0, 0, 1, 20.5, 5.5)">font_family_test</text>
</svg>
"""

svg = svglib.svg2rlg(StringIO(s))
drawToFile(svg, "font_family_test.pdf")
print(svg.asString(format="svg"))
```

giving 

```
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE svg
  PUBLIC '-//W3C//DTD SVG 1.0//EN'
  '[http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd'>](http://www.w3.org/TR/2001/REC-SVG-20010904/DTD/svg10.dtd'%3E)
<svg width="150.0" height="40.0" preserveAspectRatio="xMinYMin meet" viewBox="0 0 150 40" xmlns="http://www.w3.org/2000/svg" version="1.0" fill-rule="evenodd" xmlns:xlink="http://www.w3.org/1999/xlink">
	<title>...</title>
[...]

PR #450, opened 2026-04-28
Fix bare-number/px unit inconsistency and coordinate system (2.0b1)

### LINKED ISSUE #439: Inconsistency between bare-number and px unit handling in convertLength
This issue documents a regression introduced in #387 (v1.6.0) and the fundamental tension it surfaces.

## What changed in #387

To correctly implement the CSS/SVG spec (`1px = 1/96 inch`, `1pt = 1/72 inch`, so `1px = 0.75pt`), the `convertLength` method was updated:

- `px`: `value` → `value × 0.75`
- `pt`: `value × 1.25` → `value` (already in points, ReportLab's internal unit)

This is **correct in isolation** — ReportLab uses points, and `96px = 72pt = 1 inch`.

## The problem

The SVG spec defines the **user unit** (bare number, no unit suffix) as equivalent to `px`:

> "1 user unit = 1px"

But bare numbers in `convertLength` go through ReportLab's `toLength()`, which returns them unchanged (1:1). This means:

| Input | Result | Should be |
|-------|--------|-----------|
| `font-size="13"` | `13.0` pt | `9.75` pt |
| `font-size="13px"` | `9.75` pt | `9.75` pt ✓ |

So `font-size="13"` and `font-size="13px"` — which are spec-equivalent — now render at different sizes. This is the regression reported in #433.

## Options

1. **Revert `px` to 1:1** (back to v1.5.1 behaviour): internally consistent but spec-incorrect. `px` and `pt` are treated as equal.

2. **Scale bare numbers by 0.75 too**: spec-correct and consistent, but shifts the entire coordinate system — positions, sizes, strokes — by 0.75, a significant breaking change for existing users.

3. **Scale at the document level**: apply the px→pt factor to the SVG viewport/viewBox transform rather than to individual lengths, so the whole drawing scales uniformly. This is the most correct approach but requires a deeper architectural change.

Contributions and opinions welcome.

deeplook:
Fixes #439.

## Summary

SVG spec §5.9.2 defines 1 user unit = 1 px = 0.75 pt (at 96 dpi). Previous
releases treated 1 user unit as 1 pt, causing `font-size="13"` and
`font-size="13px"` to render at different sizes, and all output to be 33%
larger than a browser would produce.

## Changes

- `convertLength` now returns **user units (px)** instead of ReportLab points.
- New `convertLengthToPt` helper multiplies by `PX_TO_PT = 0.75` for the few
  places that need absolute points (font sizes, canvas dimensions).
- `get_box` parses `viewBox` as raw floats — never applies unit conversion
  (viewBox defines a unitless coordinate space, SVG spec §8.1).
- `renderSvg` viewport scale = `canvas_pts / viewBox_user_units`, which
  naturally incorporates ×0.75 for all shape coordinates via the group
  transform.
- SVGs with a `viewBox` but no explicit `width`/`height` now fall back to
  viewBox dimensions for the scale, fixing cropping of such files.
- `DEFAULT_FONT_SIZE` restored to `12` (pt, the CSS default); `em` bases use
  `DEFAULT_FONT_SIZE / PX_TO_PT = 16` user units.
- Regression tests added for bare-number == px consistency at both the
  `convertLength` level and the rendered `fontSize` level.

## ⚠️ Breaking change

Output PDF sizes will differ from 1.x: any SVG whose dimensions are expressed
in user units or `px` will produce a document that is 75% of its previous
linear size (correct physical size per spec).

**Migration** — to restore 1.x dimensions:

```python
drawing = svg2rlg("file.svg")
factor = 4 / 3   # 1 / 0.75
drawing.width  *= factor
drawing.height *= factor
drawing.scale(factor, factor)
```

See `CHANGELOG.rst` for full details.

## Related issues filed

- #448 — `preserveAspectRatio` not implemented (pre-existing, unrelated)
- #449 — SVG 2 viewport units `rem`/`vw`/`vh` not yet supported (future work)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,21 +3,39 @@
 ChangeLog
 =========
 
-Unreleased
-----------
+2.0b1 (2026-04-29)
+------------------
+
+**Breaking change — output sizes will differ from 1.x.**
+
+svglib now correctly maps SVG user units to ReportLab points using the
+standard SVG/CSS conversion factor: 1 px = 0.75 pt (96 dpi).  Previous
+releases treated 1 user unit as 1 pt, which is 33 % too large.  Any SVG
+whose ``width``/``height`` or ``viewBox`` uses user units or ``px`` will
+produce a PDF that is 75 % of its previous linear dimensions (same
+proportions, correct physical size).
+
+*Migration* — if you need to preserve the old apparent size, scale the
+returned ``Drawing`` object before use::
+
+    from svglib.svglib import svg2rlg
+
+    drawing = svg2rlg("file.svg")
+    # Restore 1.x dimensions (1 user unit → 1 pt, non-spec):
+    factor = 4 / 3          # 1 / 0.75
+    drawing.width  *= factor
+    drawing.height *= factor
+    drawing.scale(factor, factor)
+
 - Move ``rlpycairo`` to the ``bitmaps`` extra, so Cairo is no longer part of
   the default installation path.
 - Fix inconsistent unit handling: bare numbers and ``px`` units now produce
   identical output, as required by the SVG spec (§5.9.2). ``convertLength``
   now returns user units (px) instead of ReportLab points; a new
   ``convertLengthToPt`` helper is used where absolute point values are needed
-  (font sizes, canvas dimensions). **Output sizes will change** for SVGs that
-  mix bare-number and ``px`` lengths, or that use physical units (``cm``,
-  ``mm``, ``in``, ``pt``, ``pc``) without a ``viewBox`` — the rendered PDF
-  will now match browser output rather than the previous (incorrect) scaling
-  (#439).
+  (font sizes, canvas dimensions) (#439).
 - Fix SVGs with a ``viewBox`` but no explicit ``width``/``height``: the
-  viewport scale now defaults to PX_TO_PT (0.75) instead of 1, preventing
+  viewport scale now defaults to ``PX_TO_PT`` (0.75) instead of 1, preventing
   content from being cropped.
 - Add support for SVG ``linearGradient`` and ``radialGradient`` paint servers
   (#442).
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,18 @@ Unreleased
 ----------
 - Move ``rlpycairo`` to the ``bitmaps`` extra, so Cairo is no longer part of
   the default installation path.
+- Fix inconsistent unit handling: bare numbers and ``px`` units now produce
+  identical output, as required by the SVG spec (§5.9.2). ``convertLength``
+  now returns user units (px) instead of ReportLab points; a new
+  ``convertLengthToPt`` helper is used where absolute point values are needed
+  (font sizes, canvas dimensions). **Output sizes will change** for SVGs that
+  mix bare-number and ``px`` lengths, or that use physical units (``cm``,
+  ``mm``, ``in``, ``pt``, ``pc``) without a ``viewBox`` — the rendered PDF
+  will now match browser output rather than the previous (incorrect) scaling
+  (#439).
+- Fix SVGs with a ``viewBox`` but no explicit ``width``/``height``: the
+  viewport scale now defaults to PX_TO_PT (0.75) instead of 1, preventing
+  content from being cropped.
 - Add support for SVG ``linearGradient`` and ``radialGradient`` paint servers
   (#442).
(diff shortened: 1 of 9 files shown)


mosc9575:
I tried this changes and they solve the problem I had with the font size. The transformation from SVG to PDF works well.

But the new factor `PX_TO_PT = 0.75` is not reverted in `reportlab`.

The function `setFont` of class `SVGCanvas` in `reportlab.renderSVG` looks like 

```python
  def setFont(self, font, fontSize):
      if self._font != font or self._fontSize != fontSize:
          self._font = font
          self._fontSize = fontSize
          style = self.style
          for k in TEXT_STYLES:
              if k in style:
                  del style[k]
          svgAttrs = self.fontHacks[font] if font in self.fontHacks else {}
          if isinstance(font,RLString):
              svgAttrs.update(iter(font.svgAttrs.items()))
          if svgAttrs:
              for k,v in svgAttrs.items():
                  a = 'font-'+k
                  if a in TEXT_STYLES:
                      style[a] = v
          if 'font-family' not in style:
              style['font-family'] = font
          style['font-size'] = '%spx' % fontSize
```

In the last line the point to pixel ratio is `1:1`. This should be reported.

My initial example in #433 shows this behavior. The initial font size is `13px` when calling `svglib.svg2rlg` and after calling `svg.asString(format="svg")` the font size shrinks to `9.75px` which is odd. I know, that this is an issue in `reportlab`.

It is great the the one direction now works as expected.

Once again I want to thank you for your work.

deeplook:
Re: font size shrinkage in SVG round-trip

This is a known limitation caused by a bug in ReportLab's `renderSVG` module. When ReportLab renders a drawing back to SVG, `SVGCanvas.setFont` writes font sizes with `px` units even though the value is already in points:

```python
# reportlab/graphics/renderSVG.py
style['font-size'] = '%spx' % fontSize  # should be 'pt', not 'px'
```

Minimal reproduction with ReportLab 4.4.3:

```python
from reportlab.graphics.shapes import Drawing, String
from reportlab.graphics import renderSVG

# A string at 9.75pt (equivalent to 13px at 0.75 scale)
d = Drawing(200, 100)
d.add(String(10, 50, "Hello", fontSize=9.75))

svg_out = renderSVG.drawToString(d)
print(svg_out.decode())
# Output contains: font-size: 9.75px
# But 9.75 is in points — should be font-size: 9.75pt (or font-size: 13px)
```

The fix in ReportLab would be to change the last line of `SVGCanvas.setFont` to:

```python
style['font-size'] = '%spt' % fontSize
```

This is outside svglib's control. The `svg2rlg` direction (SVG→RLG) now works correctly — the round-trip shrinkage is a pre-existing ReportLab issue.

mosc9575:
@deeplook Thanks for you summary. I will send a mail to a reportlab author and link your comment to fix also this issue.

claudep:
@deeplook, I'm sorry, but I will clearly be less active on this module, so don't count on me for reviews in the future. Thanks a LOT for all you are doing :medal_sports:

deeplook:
> @deeplook, I'm sorry, but I will clearly be less active on this module, so don't count on me for reviews in the future. Thanks a LOT for all you are doing 🏅

@claudep Thank _you_ for your many contributions in the past! Wishing you all the best!

replabrobin:
Not sure exactly how to improve the situation in the general case. I propose a small change in reportlab's renderSVG.py that adds an extra keyword argument fontSizer to SVGCanvas. If it's a string (default 'px') then that is used  as a unit in the  output. So 3 --> 3px. If fontSizer is given as a callable then it would be used to obtain the string value for the font-size attribute.  Given the larger changes in svglib I'm uncertain if that's enough or even needed. Reportlab assumed 1px = 1pt a long time ago, but that's purely for convenience and in practice sizes are adjusted to match the output canvas. So the PMCanvas has a dpi argument.

deeplook:
Robin, thanks for joining the discussion! The fontSizer hook direction could make sense, especially since renderSVG is a general-purpose exporter and a hard px→pt swap would be a silent breaking change for ReportLab users in general.

For the svglib round-trip case specifically: after PR #450, font sizes flowing through svglib are stored in points internally. So what is needed from ReportLab is a way to emit pt units (or an explicit numeric conversion) when writing font-size in SVGCanvas.setFont. A callable like fontSizer=lambda v: f'{v}pt', or a simpler font_size_unit='pt' string kwarg with 'px' as the default, that would let svglib opt into correct round-trip behavior without affecting existing ReportLab→SVG consumers.

Both should work for svglib. I'm happy to test a patch once there's something to try. Would that work?

PR #454, opened 2026-05-16
Default nested <svg> width/height to parent viewport per SVG1.1 §5.1.2

njacquemin1993:
## Problem

When a nested `<svg>` element has a `viewBox` but omits `width` / `height`,
`renderSvg` leaves both at `None` and skips the viewBox-to-viewport scaling
entirely. Inner content then renders at viewBox-unit size instead of filling
the parent slot.

Per [SVG 1.1 §5.1.2](https://www.w3.org/TR/SVG11/struct.html#SVGElementWidthAttribute),
those attributes default to `100%` of the parent viewport, so the scaling
*should* be applied — the parent's viewport size mapped onto the nested
viewBox.

(SVG 2 actually changed this rule so that omitting either resolves to 0
and disables rendering, but no browser implements that and most real-world
SVGs follow SVG 1.1. svglib isn't trying to support SVG 2 yet, so the SVG 1.1
behavior is what matches user expectation.)

### Real-world impact

This is what made me hit it: the [`chqr`](https://pypi.org/project/chqr/)
package (Swiss QR-bills) embeds the QR code as a nested
`<svg viewBox="0 0 N N">` with no `width` / `height`. With the current code,
the QR comes out at ~25 % of its intended 46 mm size, regardless of any
outer scaling.

## Fix

In the `elif view_box:` branch of `renderSvg`, fall back to `_saved_box`
(the parent viewport, already captured at the top of the method) when
`width` or `height` is `None` and we're not at the outermost svg.

```python
if not outermost and _saved_box is not None:
    if width is None:
        width = _saved_box.width
    if height is None:
        height = _saved_box.height
```

This is safe by construction:

Outermost svg — branch is skipped, behavior unchanged.
Nested svg with explicit width / height — fallback is skipped, behavior unchanged.
Nested svg with viewBox but no width / height (the broken case) — now scales to fill the parent viewport.
The existing TestEmbedded::test_svg_in_svg and test_embedded_svg_with_percentages continue to pass.

## Test
Added TestEmbedded::test_nested_svg_without_width_height asserting
scale = 4 for a 50×50 viewBox inside a 200×200 parent.

Full unit-test suite (tests/test_basic.py + tests/test_fonts.py):
112 passed, no regressions.

### DIFF
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -1496,6 +1496,16 @@ class SvgRenderer:
             width, height = self.shape_converter.convert_length_attrs(
                 node, "width", "height", defaults=(None,) * 2
             )
+            # Per SVG2 §8.2, a nested <svg> with omitted width/height defaults
+            # them to 100% of the parent viewport. Without this fallback, an
+            # inner <svg viewBox="0 0 N N"> with no explicit width/height
+            # renders content at viewBox-unit size instead of filling the
+            # parent slot.
+            if not outermost and _saved_box is not None:
+                if width is None:
+                    width = _saved_box.width
+                if height is None:
+                    height = _saved_box.height
             if height is not None and view_box.height != height:
                 y_scale = height / view_box.height
             if width is not None and view_box.width != width:
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -1496,11 +1496,14 @@ class SvgRenderer:
             width, height = self.shape_converter.convert_length_attrs(
                 node, "width", "height", defaults=(None,) * 2
             )
-            # Per SVG2 §8.2, a nested <svg> with omitted width/height defaults
-            # them to 100% of the parent viewport. Without this fallback, an
-            # inner <svg viewBox="0 0 N N"> with no explicit width/height
-            # renders content at viewBox-unit size instead of filling the
-            # parent slot.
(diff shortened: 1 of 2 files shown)


deeplook (commented):
Thanks for the fix and the real-world motivating case!

One small correction: the `100%` default for nested `<svg>` width/height seems to come from **SVG 1.1**, not SVG 2 — SVG 2 actually changed this so that omitting `width`/`height` on a nested `<svg>` resolves to `0` (disabling rendering). That said, the fix is correct in practice: no browser implements the SVG 2 behavior, most real-world SVGs follow SVG 1.1 conventions, and svglib isn't trying to support SVG 2 yet (that would likely be a larger effort). Could you update the code comment and PR description to cite SVG 1.1 instead?

The test assertion on `transform[0]` / `transform[3]` is a bit fragile — if ReportLab's transform representation changes, the failure message won't be obvious. Consider asserting on the rendered rect's actual dimensions instead, similar to how `test_embedded_svg_with_percentages` checks `inner_rect.width == 15`.

Minor points aside, the logic is sound. Happy to merge once those are addressed.

njacquemin1993:
Thanks for the feedback @deeplook !
Your points are now addressed

deeplook:
@njacquemin1993 Thank you!!

PR #470, opened 2026-06-26
Issue #366: Fix missing transform on clip paths

blayzen-w:
See issue #366

I attempted to fix this a while back with pull request #367 but I abandoned the work after finding that transforms were not supported on clip paths within pdfs.

This pull request takes a different approach by applying the transform directly to the clip path shape points.

Let me know if you would like anything changed.

## The issue

When creating a clipping path the transform is discarded in the `get_shape_from _node` function. The `path` case has the transform applied to the group, but the group is discarded and only the child is returned. The `rect` case uses [convertRect](https://github.com/deeplook/svglib/blob/28a2513dab4f092f93d0155d600fee5355bf2dc3/src/svglib/svglib.py#L1857) which does not use the transform.

https://github.com/deeplook/svglib/blob/28a2513dab4f092f93d0155d600fee5355bf2dc3/src/svglib/svglib.py#L1340-L1352

This causes an issue with cases like in this svg where a transform is applied to the clip path shape.

<svg namespace="http://www.w3.org/XML/1998/namespace" 
     xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" 
     viewBox="0 0 100 50" width="100" height="50">

    <defs>
        <clipPath id="clip-1">
            <rect transform="scale(0.5, 0.5)" height="100" width="100" x="0" y="0"/>
        </clipPath>
    </defs>
    <g>
        <g clip-path="url(#clip-1)">
            <path fill="green" d="M 0 0 H 50 V 50 H 0 Z"/>
            <path fill="red" d="M 50 0 H 100 V 50 H 50 Z"/>
        </g>
    </g>
</svg>

The clip path should be scale down by 50%. In the svg you will only see the green square. But in the generated pdf you will see both green and red squares.

Based on my previous PR, I do not think that pdf allow for transforms on the clipping path shape. This creates a bit of a issue because svgs can apply transforms that cannot be represented by the same type of shape element. For example rotating a rectangle defined by x, y, width and height by 45 deg.

## The fix

This pull request attempts to fix the issue by converting any non-path shapes to paths, and applying the transform to the x,y point pairs in the path. Rounded corner rectangles are a bit tricky and have to be approximated with a bezier curve. Other shape elements, like ellipses and circles, could be supported using a similar approach by creating a close approximation with bezier curves but I felt like that was out of the scope of this change.


deeplook:
@blayzen-w Sorry but this PR has slipped past my attention before. Thanks a lot for tackling it, and for the clear write-up!

I really like the approach of baking the child transform directly into the path coordinates. It sidesteps a lot of awkwardness, and the rect to path conversion is a nice pragmatic approach.

Apparently, one part of the underlying problem actually might be on the ReportLab side. Its PDF renderer deliberately doesn't save/restore the graphics state around clip paths (so the clip persists to following siblings), which means any transform on a clip path leaks into the subsequently drawn content rather than being confined to the clip geometry.

So in some sense this PR can be seen as a workaround for that, it keeps everything self-contained in svglib and avoids the inverse-matrix headaches a native fix would involve. I won't go into all the details here, but I've just pinged Robin at ReportLab about it to see what they think.

It would be great to eventually extend this to the other valid clip shapes too. polygon/polyline should be cheap, and circle/ellipse can reuse the Bézier pattern you already worked out for rounded corners. Text clipping could be the tricky one and maybe would be the real challenge, maybe for another day.

In any case, this would be a really nice feature for the 2.1 release!

blayzen-w:
@deeplook Thanks for taking a look at the changes. I took another look at the graphical state issue within pdfs and I looks like the inverse transform you mentioned would solve  the issues in the pdf.

Take the following pdf that mirrors the test case but with a clipping mask size of 95x95. The smaller clip mask size allows you to see if the clipping mask is working correctly or not.

```
%PDF-1.4

1 0 obj
<< /Type /Catalog /Pages 2 0 R >>
endobj

2 0 obj
<< /Type /Pages /Count 1 /Kids [3 0 R] >>
endobj

3 0 obj
<<
/Type /Page
/Parent 2 0 R
/MediaBox [0 0 75 37.5]
/Resources <<
  /Font << /F1 4 0 R >>
>>
/Contents 5 0 R
>>
endobj

4 0 obj
<<
/Type /Font
/Subtype /Type1
/BaseFont /Times-Roman
>>
endobj

5 0 obj
<< /Length 382 >>
stream

BT
/F1 12 Tf
14.4 TL
ET

0 0 0 RG
1 w
0 J
0 j
[] 0 d
0 0 0 rg

BT
/F1 10 Tf
12 TL
ET

.75 0 0 -0.75 0 37.5 cm

% clipping mask
0 0 95 95 re
W
n

% green square
0 .501961 0 rg
0 0 50 50 re
f

% red square
1 0 0 rg
50 0 50 50 re
f

endstream
endobj

xref
0 6
0000000000 65535 f
0000000010 00000 n
0000000065 00000 n
0000000120 00000 n
0000000225 00000 n
0000000310 00000 n

trailer
<<
/Root 1 0 R
/Size 6
>>
startxref
420
%%EOF
```

If you wrap the clipping mask with a q/Q and a 50% scale transform `0.5 0 0 0.5 0 0 cm` to get the following:
```
q
  0.5 0 0 0.5 0 0 cm
  
  % clipping mask
  0 0 95 95 re
  W
  n
Q
```

Then the clipping mask disappears. I'm assuming this is because clipping paths are stored in the current graphical state. When we wrap the clip path in a new graphical state to apply the transform, we also remove it from the graphic state that it should be clipping.

However, if you apply the inverse transform to the contents being clipped and move it into the clipping mask state, then you get the correct results. The clipping mask in this case is made larger to make sure its working.

```
q
  0.5 0 0 0.5 0 0 cm

  % clipping mask
  0 0 105 105 re
  W
  n

  q
    % inverse mask transform
    2 0 0 2 0 0 cm

    % green square
    0 .501961 0 rg
    0 0 50 50 re
    f

    % red square
    1 0 0 rg
    50 0 50 50 re
    f
  Q
Q
```

If Reportlab could fix that issue then I think keeping the transforms would be a better approach. This change could bridge the gap in the mean time unless you know of a easier way to achieve this result. I sent them a email the last time I was working on this but never heard back so I assumed it was a dead end.

blayzen-w:
While I was testing the changes I noticed that ReportLab does not draw rounded corner rectangles correctly with pdfs. They seem to ignore the `ry` value and only use `rx` for a single radius value.  This can be seen in `reportlab.graphics.renderPDF.py` `_PDFRenderer.drawRect`, which eventually calls `reportlab.pdfgen.pathobject.py` `PDFPathObject.roundRect` for the drawing.

They are using a similar bezier approximation but with a radius constant of 0.4472 instead of 0.5522. Their value has a much higher error rate with the approximation and is visually different than the svg rounded corner. According to [this page](https://spencermortensen.com/articles/bezier-circle/) the 0.5522 number has a drift of only ~0.027%.

Do you think we should try to monkey patch this in report lab, or convert all rounded corner rectangles into paths using `convertRectPath`? I can open up a different pr if needed.

deeplook:
Frankly, I haven't looked at the generated PDF at all...

I'm pasting here a response I got from Robin regarding transforms. It sounds much like early design decisions made in reportlab, that would require some substantial effort to change, and my impression is that ReportLab, the company, might not have an urgent immediate interest in doing so.

> Robin:
> The problem is that of different graphics models. So far as I understand it the CTM changes are divorced from the 'node' so in reportlab I don't think there is a clipping path node where the CTM could be attached. I think the analysis is correct in that drawPath does not save/restore state; the reportlab graphics models assume that state is on the canvas (I think Andy modeled this on PDF). The intention is that transforms are applied to the canvas (itself a source of huge confusion since everything seems backwards).
> 
> I'm not sure where the bit about apply state changes comes from. Since the generic reportlab node does not have transforms (supposedly only groups and drawings do) it must be an attribute abuse that allows a Path object to have one. The code in reportlab does not prevent Path objects from having a transform attribute and that will be used as it passes through the render walk; I'm not sure if the svglib ClippingPath allows transforms, but if it does, they get set on the reportlab Path node and will be applied as the node goes through.
> 
> A secondary issue is the path model itself. Reportlab paths are not any kind of container so there's no 'content'. As the AI notes it's the siblings that are affected by clips.
> 
> As for the reportlab 'remedy' it's unclear where reportlab should apply a saveState and restoreState and under what conditions. I suspect it's more easily done in svglib.
> 
> The obvious attempt to wrap the clip in its own Group with a transform does not seem to work for me.

So we might want to aim for a fix in svglib only. I'll paste a more detailed analysis that I've tasked either Claude or Codex with below.

deeplook:
Ok, this is what Codex or Claude had to say (I really forgot which one, since I use both):

# Transforms on clipPath Children: Analysis of Issue #366 / PR #470

## The Bug

When an SVG `<clipPath>` contains a child element (e.g. `<rect>` or `<path>`) with a `transform` attribute, svglib silently discards that transform. The function `get_shape_from_node` (svglib.py, around line 1340) extracts the raw shape from the child element but never applies the child's transform to it. For `<path>` children, `convertShape` wraps the shape in a `Group` to carry the transform, but the code discards the group and returns only the inner shape. For `<rect>` children, `convertRect` ignores the `transform` attribute entirely.

The result: the clipping region is computed from the untransformed geometry, so a scaled, rotated, or translated clip shape is applied as if no transform existed.

## What the Spec Says

SVG 1.1 section 14.3 defines the clipping model as:

> "The **raw geometry** of each child element **exclusive of rendering properties such as `fill`, `stroke`, `stroke-width`** within a `clipPath` conceptually defines a 1-bit mask."

Transforms are not rendering properties — they define an element's coordinate system and are therefore part of its geometry. Transforms on `clipPath` children are valid per spec and must be respected when computing
 the clip region. All major browsers implement this correctly.

The relevant authoritative spec today is CSS Masking Module Level 1 (clipPath was moved out of SVG into that module). It inherits the same "raw geometry" model from SVG 1.1 and does not alter the semantics around child transforms.

SVG 1.2 Tiny (the only completed 1.2 spec) dropped `clipPath` entirely. SVG 2 remains an unfinished Candidate Recommendation (last published October 2018, never advanced to Recommendation). Neither changes the analysis.

## Does PDF Lack the Capability?

No. PDF can handle this natively. The `cm` operator modifies the Current Transformation Matrix (CTM), and the `W`/`W*` (clip) operators record the clip in device space at the moment they are called. The valid PDF sequence is:

```
q                         % save graphics state
a b c d e f cm           % apply child element's transform
... path construction ... % define clip shape in transformed space
W n                       % set clip (recorded in device space)
ia ib ic id ie if cm     % apply inverse transform to restore CTM
... draw content ...      % content drawn in original space, clipped correctly
Q                         % restore graphics state
```

## Why ReportLab Is the Right Place for the Native Fix

ReportLab's `renderPDF.drawNode` deliberately skips `saveState`/`restoreState` for `ClippingPath` nodes so that the clip persists into subsequent sibling rendering. However, this means any CTM change from the ClippingPath's own `transform` also persists and leaks into the content drawing, scaling or rotating the content incorrectly.

A fix in ReportLab's renderer would be: when a `ClippingPath` node has a `transform`, emit the transform via `cm`, draw the clip, then emit the inverse transform via `cm` to restore the CTM before content is drawn. ReportLab already applies transforms for ClippingPaths through `applyStateChanges`; it simply never undoes them.

## Would a ReportLab Fix Be a Complete Solution?

Not entirely. Even with a corrected ReportLab renderer, two problems remain:

1. **Inverse transform computation**: Computing the inverse of an arbitrary affine matrix requires care around degenerate or near-singular cases. A robust implementation is non-trivial.

2. **Non-path shapes with non-affine-representable transforms**: A `<rect>` rotated 45° cannot be expressed
 as a rect in any coordinate frame — it must be converted to a path first. The ReportLab fix would still require svglib to convert non-path clipPath children to paths before attaching the transform. So svglib still needs its own conversion logic regardless.

A complete native solution therefore requires changes in both ReportLab and svglib, and the coordinate-baking approach in PR #470 is simpler and avoids the inverse-transform problem entirely.

## Assessment of PR #470

The PR's approach — converting clipPath child shapes to paths and baking the transform directly into the path point coordinates — is the correct pragmatic fix for svglib. It:

- Is self-contained within svglib, requiring no ReportLab changes.
- Works for all affine transforms without needing to compute inverses.
- Handles the necessary `<rect>` → `<path>` conversion (including rounded corners approximated with Bézier curves).
- Avoids CTM pollution in the PDF output.

### What the PR Still Needs to Be Complete

1. **Ellipses and circles**: The PR author acknowledges these are out of scope but they are valid `clipPath` children per spec. A complete fix needs Bézier approximations for `<circle>` and `<ellipse>` as well, using the same coordinate-baking approach.

2. **Other valid clipPath children**: The SVG spec also permits `<polygon>`, `<polyline>`, `<line>`, `<text>`, and `<use>` as clipPath children. `<polygon>` and `<polyline>` are the most practically significant omissions — they map trivially to a path (line segments only, no Bézier needed). `<circle>` and `<ellipse>` require the same Bézier approximation already used for rounded rects. `<text>` (clip to glyph outlines) is the hardest and least common. `<use>` is partially handled already via `renderUse`.

3. **Nested transforms**: If a `<clipPath>` has a `transform` attribute on the element itself (in addition to transforms on children), and `clipPathUnits="objectBoundingBox"` is used, the coordinate spaces become more complex. The PR should be validated against these combinations.

4. **Tests for the new shape conversions**: The rounded-corner rect Bézier approximation and any future ellipse/circle approximation should have their own unit tests verifying geometric accuracy, not just visual pass/fail integration tests.

5. **Validation against complex SVGs**: The PR author notes they are still validating against larger SVGs. This should be completed before merging.

## Where Is It Less Effort to Implement?

In svglib, clearly.

- `<polygon>` and `<polyline>` are just sequences of `lineTo` calls — a handful of lines in `get_shape_from
_node` alongside the existing `rect` and `path` cases.
- `<circle>` and `<ellipse>` need Bézier approximations, but the PR already worked out that pattern for rounded rects — it is copy-and-adapt work.
- The transform-baking machinery is already implemented in the PR; new shape types just feed into it.

The ReportLab fix by contrast requires modifying the renderer's save/restore logic (with subtle correctness implications for all clip path rendering), computing matrix inverses robustly, and upstreaming a change to a separate project with its own review process. Crucially, it does not eliminate the svglib work: ReportLab only handles `Path`/`ClippingPath` objects and has no concept of `<rect>`, `<circle>`, or `<ellipse>` as clip shapes, so svglib still needs the shape-to-path conversions regardless.

The ReportLab fix is architecturally purer but the net savings in svglib effort are small and the cost is high. The right approach is to complete the fix in svglib.

blayzen-w:
Thanks for look into it and reaching out to ReportLab. It looks like baking in the transform is going to be the better approach.

I tested quite a few svgs and everything seem to work correctly so this PR should be good to go now.

I'm upgrading from a modified version of 0.9 so there was a lot to go over and test. We were stuck on on that version due to the clipping mask and text sizing issues.

I would still like to defer adding support for new shapes if possible. My main goal was to get the existing shapes working correctly before adding new features for the other shapes. I can spend some time on those after this pr gets merged in.

What do you think about the the issue with the incorrect rounded corner rendering? I can open a new issue showing the problem if needed.

deeplook:
A separate issue for the corners would make it easier to reference, but that one sounds even more like it's an issue in reportlab, and I know it's even harder now to open issues there, with the mailinglist out of service and the code in a Mercurial repo. No response from Robin on these corners, yet, but it sounds like this would be easier to fix on their side.

The clipath transforms, ok let's start simple. Tests are important. Maybe you can compare locally with cairosvg? Thanks for driving this!

blayzen-w:
What kind of test for were you thinking for cairosvg?

I could probably convert the svg into a raster image using both `drawing_from_svg` and `cairosvg.svg2png`, then compare the output using pillow on a per pixel basis. That might allow us to compare more types of shapes than just the three squares I have, and work independent of the implementation.

I originally had three test cases verifying the path data of each clipping path being tested, but a lot of the tests used the similar code and I didn't want to worry about rounding issues. So I merged them into one and just checked the max value of all points since all three shapes should go from 100px to 50px.

I've been visually checking all of my tests using a combination of GNOME Evince, Inkscape, and Illustrator.

deeplook:
Great if you already compare the output with multiple other renderers! I thought cairosvg output as a reference to compare with, on a raster-basis as you describe (surely not to run cairosvg in the test-suite). It's likely not going to be pixel-perfect, but within some small margin would do. Maybe it would be more of an exercise. This would be ideal for using many examples, but if we have only 1-2 and it is dead simple to do an eyeball test that would be better then nothing. The Wikipedia flags in the test-suite are a little bit like this, although it's not always dead-simple to see there, and they are arguably too many for an eyeball test...

blayzen-w:
I added another test case that will compare the raster output from the svg drawing against exports of the same svg from Inkscape. I tried cairosvg but there was a larger difference in the gradients with curved corners.

Do you think these new tests will work?

src/svglib/svglib.py
blayzen-w: It looks like non-transformed shapes can be found in the `get_shape_from_group`  function and slip though here. This shape conversion also doesn't seem to handle the rx or ry in the rectangle. I'll have to rethink this part and create test cases for it.

blayzen-w:
I ended up adding support for other shapes. Use tags load in already parsed shapes that could bypass the transform baking. So I changed the functionality to convert any existing shape into a path.

I didn't add Polyline since it is not considered a `SolidShape` and didn't pass the existing `isinstance(elem, SolidShape)` check. It technically works in the browser and Inkscape so I think its possible if we close the path. Do you think we should add this one?

deeplook:
Nice, and very promising! I have only a few things to add...

1. Polygon missing `moveTo` (likely a bug in svglib.py:2436)
```python
elif isinstance(shape, Polygon):
    for i in range(0, len(shape.points), 2):
        path.lineTo(shape.points[i], shape.points[i + 1])  # ← first should be moveTo
```
All other shape conversions start with `path.moveTo(...)`. Starting with `lineTo` on an empty path is undefined behaviour in reportlab and will likely emit an implicit `moveTo(0,0)` first, producing a spurious edge from the origin.

2. Variable shadowing of the Bezier constant c (fragile)
```python
c = 0.552284749830793   # ← Bezier constant
...
if transform:
    a, b, c, d, e, f = transform  # ← shadows c with the x-shear matrix element
```
This only works because the Bézier code runs before the transform block. If the method is reorganised (e.g. circles/ellipses move below the transform block), the wrong value will be used silently. Rename either one.

3. `shapeToPath` transform parameter type inconsistency
The method is called from two places with incompatible types:
- `get_shape_from_group` passes `group.transform` → a numeric 6-tuple `(a,b,c,d,e,f)`
- `get_shape_from_node` passes `child.getAttribute("transform")` → a string or None

The string case is dead code in practice (because `convertShape` always wraps a transform-bearing shape in a `Group`, so `isinstance(item, Group)` fires first), but the API is misleading and would silently crash if that path were ever reached. The signature should be typed to accept only a tuple/matrix, not a raw attribute string.

Type annotations would catch this, but they are not used everywhere, yet, and not enforced, either. That's something that should have landed for 2.0, but sadly didn't.

4. Duplicate entry in visual test list (test_basic.py:396)
```python
test_files = [
    "path", "rect", "rounded-rect", "path-no-transform",
    "rect",   # ← duplicate; "rounded-rect" is the missing entry
    "ellipse", "circle", "polygon",
]
```
"rect" appears twice; "rounded-rect" appears only in the unit-test list above but not here in the visual test list.

Minor Notes

- The use reference path relies on `get_shape_from_group` returning the first `SolidShape` found which can miss multi-shape use references, but this was already the pre-existing limitation.
- I guess `polyline` is intentionally omitted because it's not a SolidShape, which sounds reasonable, right?
- The visual test tolerance of `==0` pixel difference seems very strict (with a per-channel threshold of 3) but apparently it works given the simple SVG test cases.

tests/utils.py
```diff
@@ -15,3 +19,74 @@ def drawing_from_svg(content: str) -> Any:
 def minimal_svg_node(content: str) -> NodeTracker:
     """Convert a minimal SVG snippet to a NodeTracker."""
     return NodeTracker(XML(content), None, 0, None, False)
+
+
+def svg_raster_difference(svg: str, raster: str, scale=1, tolerance=3) -> float:
```
deeplook: Maybe prefer `svg_path` and `raster_path` as clearer parameter names?

blayzen-w:
Thanks for the review! Those issues should be fixed now.

As for the multi-shape case, I was looking into that on Thursday and I think I found a solution for it. I tried adding multiple `ClippingPath` elements to a group but Reportlab seemed to ignore them. However, since we now have support to convert any shape to a path and baking in transforms, we might be able to combine the paths into a single one using the move to command before combining the draw commands. I was planning on opening up a new PR for this once since it might get tricky with mixed fill rules. Nested groups might also be a solution to explore.

I made the visual tests have a tight tolerance so it could catch slight differences in the paths, like the 0.4472 bezier constant issue in Reportlab. That change results in a 2.26% difference between the images. The pixel color distance tolerance should eliminate most or all of the nose from differences in gradients, but I can open this up a slight bit if you want.

deeplook:
Looks great to me! I'd suggest to merge it as-is, and take a few days to test it, while you can think about a follow-up PR if you want and/or add yourself to the contributors file. I'll make the additions to the changelog later. I'd aim to release this in 2.1.0, including some added type-safety features, which I haven't started, yet. Let me know if that's ok for you!

blayzen-w:
That works for me.

I'll follow up with a separate PR for the multiple clip paths. It looking like its going to be difficult with the Reportlab limitations.

PR #483, opened 2026-07-09
Add image size fallback if the width/height attributes are missing (fixes #474)

### LINKED ISSUE #474: Embedded images without a defined width/height do not fall back to the size of the image
Embedded images are sized based on the width and height attributes. When these are not provided the image is scaled to a size of 0, 0 which creates a transform that cannot be inverted, resulting in a crash when exporting to a png. The image also disappears in a pdf export.

https://github.com/deeplook/svglib/blob/93713f452ba269d4858245ee1388a83ef8374107/src/svglib/svglib.py#L2219-L2230

It looks like the image has a size property with the width and height that could be used as a fallback when the attributes are not set.

This can be reproduced with the following svg:
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" height="100" version="1.1" viewBox="0,0,100,100" width="100">
    <image id="image-1"
               xlink:href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAGQAAABkCAYAAABw4pVUAAAACXBIWXMAAAfPAAAHzwGGUVlPAAAAGXRFWHRTb2Z0d2FyZQB3d3cuaW5rc2NhcGUub3Jnm+48GgAAAQxJREFUeJzt3LENwEAIwEA+++/8qTNBLHQ3Acii5dyZOwucFVvMPH8PwJcgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSMyZu+Oj3BYuJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJOYFi6UFwwW3DeAAAAAASUVORK5CYII="/>
</svg>
```

I should be able to work on this later unless anyone else wants to grab it.

blayzen-w:
This pull request adds support for falling back to the image width/height if the attributes or styling do not define it. It will preserve the image aspect ratio in the case that only one attribute is set.

Closes #474

### DIFF
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -2353,6 +2353,21 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
             node, "x", "y", "width", "height"
         )
         image = node._resolved_target
+
+        # If the width and/or height were not set then the image size
+        # needs to be used as a fallback
+        if not width or not height and image and image.size:
+            image_width, image_height = image.size
+
+            if image_width and image_height:
+                if width and not height:
+                    height = width * image_height / image_width
+                elif height and not width:
+                    width = height * image_width / image_height
+                else:
+                    width = image_width
+                    height = image_height
+
         image = Image(int(x), int(y + height), int(width), int(height), image)
 
         group = Group(image)
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -2356,7 +2356,7 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
 
         # If the width and/or height were not set then the image size
         # needs to be used as a fallback
-        if not width or not height and image and image.size:
+        if (not width or not height) and image and image.size:
             image_width, image_height = image.size
 
             if image_width and image_height:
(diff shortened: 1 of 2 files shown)


src/svglib/svglib.py
```diff
@@ -2353,6 +2353,21 @@ def convertImage(self, node: Any) -> Any:
             node, "x", "y", "width", "height"
         )
         image = node._resolved_target
+
+        # If the width and/or height were not set then the image size
+        # needs to be used as a fallback
+        if not width or not height and image and image.size:
```
deeplook:
```suggestion
        if (not width or not height) and image and image.size:
```
Small precedence nit: as written this parses as `(not width) or (not height and image and image.size)`, so the `image`/`image.size` guard only covers the height branch. It's harmless today since `image` is always a resolved PIL image here, but the parens match the intent and are more robust. Nice fix otherwise!

blayzen-w:
Nice catch on the parentheses, that should be fixed now

PR #484, opened 2026-07-09
Preserve fill rule in clipping paths

blayzen-w:
This PR will preserve the fill rule and other shape properties in clipping paths. Tests now verify the correct values.

This should mirror the old behavior in the rectangle clipping path found here:
https://github.com/deeplook/svglib/blame/9be5dac1591d6411c92cac761ea0fb2c922cd2e4/src/svglib/svglib.py#L1389-L1400

### DIFF
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -2648,6 +2648,9 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
             # Unsupported shape conversion
             return None
 
+        # Copy over fill rule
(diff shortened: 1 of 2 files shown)


deeplook:
LGTM — clean fix, and nice tests across all the shape types. Checked that it only touches the two clipping-path call sites, so no wider impact.

Tiny nit: the `# Copy over fill rule` comment is a bit narrow — `copy_shape_properties` actually copies all properties (harmless here). Maybe `# Preserve fill rule and other shape properties`. Non-blocking.

Issue #485, opened 2026-07-09
Clip rule attribute is not supported and fill rule is used instead


blayzen-w:
It looks like the fill-rule is being parsed and used in place of the clip-rule within clipping paths. The clip-rule attribute is currently being ignored.

This can be seen with the following svg:
```
<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 110 90" width="110" height="90">
    <defs>
        <clipPath id="clip-1">
            <polygon
                clip-rule="evenodd"
                points="50,0 21,90 98,35 2,35 79,90" />
        </clipPath>
    </defs>
    <g>
        <rect height="90" width="110" x="0" y="0"/>
        <g clip-path="url(#clip-1)">
            <rect fill="green" x="0" y="0" width="110" height="90"/>
        </g>
    </g>
</svg>
```

I didn't even realize that clip paths used a different property name for the fill rule when the path is inside of a clipping path until just now. The multiple clipping path shape feature has been one deep rabbit hole.

I think it would be fairly easy to add support for the property under the fill rule here:
https://github.com/deeplook/svglib/blob/c82275981f50d8c2d3d6aa5d1118fd936d759047/src/svglib/svglib.py#L2434

But I believe that would incorrectly allow both the fill-rule and clip-rule to be parsed on all elements. If the behavior is corrected where fill-rule no longer applies to the clip path and only clip-rule works, then it could break things for people who rely on the incorrect behavior.

The use case also looks even more difficult to correctly parse.
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" viewBox="0 0 110 90" width="110" height="90">
    <defs>
        <polygon id="star" clip-rule="evenodd" points="50,0 21,90 98,35 2,35 79,90" />
        <clipPath id="clip-1">
            <use xlink:href="#star"/>
        </clipPath>
    </defs>
    <g>
        <rect height="90" width="110" x="0" y="0"/>
        <g clip-path="url(#clip-1)">
            <rect fill="green" x="0" y="0" width="110" height="90"/>
        </g>
    </g>
</svg>
```

@deeplook What are you thoughts on this property and the feasibility of conditionally parsing the attribute depending on the parent/context?

deeplook:
[...]

PR #486, opened 2026-07-13
Support multiple clipping path shapes

blayzen-w:
This pull requests add support for multiple clipping path shapes by combining the shape paths into one. Additionally this fixes issue #485 by parsing the clip-rule instead of fill-rule with clipping paths.

Mixed clip-rules are not currently supported.

It looks like trying to support mixed clip rules is going to be quite difficult. Inkscape is unable to correctly handle it and just combines the shapes ignoring the clip rule, like in this PR. Illustrator can handle it correctly, but from what I can tell it's converting the self intersecting paths into independent shapes. I think it might be better to try and handle this case in a different PR since it will probably require adding support for boolean operations on shapes which is non-trivial for bezier curves.


deeplook:
Nice work on this — the multi-shape clipping is a really solid addition, and the tests are great. Honoring `clip-rule` for the direct clip children is exactly right too. This looks good to merge as a stopgap for #485 rather than sitting on it waiting for the fuller fix.

One thing worth calling out first: this picks up `clip-rule` from the clip child (and from a `<use>` itself), but not from the *thing a `<use>` points at*. So that second SVG in #485 — where `clip-rule="evenodd"` lives on the referenced `<polygon id="star">` — still comes out as `nonzero`. That's the case that really needs the "this is a clip" flag pushed down into `applyStyleOnShape` ("option 2"), so no worries leaving it out here — just worth writing down so #485 doesn't look fully done when it isn't.

Couple of tiny things:
- Drop a one-line comment in the `use` branch of `find_shapes_from_node` saying `clip-rule` on the referenced element isn't handled yet, so the gap is documented right where it is.
- Maybe make `"evenodd" in clip_rule_value` an exact match (`== "evenodd"`)? Works fine now, just a substring check that could trip things up down the line.

README note, the `xfail` for mixed rules, and the sample images all look good. With those two tweaks it can go in, and #485 stays open with a follow-up note.

blayzen-w:
I was using the substring match for the evenodd value for when `!important` is used, but I can see how invalid text could incorrectly match. I'll follow the `applyStyleOnShape` functionality and remove the `!important` string and strip out white space. I haven't had much time to work on this recently, but I should be able to get to it sometime next week.

blayzen-w:
I got those changes finished. I believe it should be good to go now.

deeplook:
Validated locally with a real renderPM backend — clipping visual tests pass, `test_mixed_fill_rules` correctly xfails, full suite green, clean under `mypy --strict` and `ruff`. The `clip-rule` fix is spec-correct and the combine approach is sensible. 👍

Two trivial nits: typo in the new comment (`lip-rule` → `clip-rule`), and worth a changelog note that `fill-rule`-in-`clipPath` behavior changes. Happy to handle both post-merge if you'd rather not respin the branch.

blayzen-w:
I added a note to the fixed section in the changelog about the clip-rule support/change and fixed the typo. I think that was the correct location for the note, but if not feel free to modify it.

deeplook:
Great, thanks! Will create a 2.1 release once #494 is ready.

PR #488, opened 2026-07-13
Images with a transform are not placed correctly (fixed #487)

### LINKED ISSUE #487: Images with a transform are not placed correctly
Take the following contrived svg test case:
```
<svg version="1.1" viewBox="0 0 50 50" width="50" height="50" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
  <image x="0" y="0" width="100" height="100" transform="scale(0.5, 0.5)" 
   xlink:href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAGQAAABkCAYAAABw4pVUAAAACXBIWXMAAAfPAAAHzwGGUVlPAAAAGXRFWHRTb2Z0d2FyZQB3d3cuaW5rc2NhcGUub3Jnm+48GgAAAQxJREFUeJzt3LENwEAIwEA+++/8qTNBLHQ3Acii5dyZOwucFVvMPH8PwJcgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSIwgMYLECBIjSMyZu+Oj3BYuJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJEaQGEFiBIkRJOYFi6UFwwW3DeAAAAAASUVORK5CYII="/>
  <rect width="25" height="50" x="0" y="0" transform="scale(0.5, 0.5)" fill="#0000ff" />
</svg>
```

It contains a image with a green background and a red square in the top right corner. The blue rectangle should align with the red square in the top right corner.

The svg should produce the following ouput.

Image

However the image is incorrectly placed and only the blue square is visible. The image location should be at 0,0 but it is instead placed at 0,100 which is outside the view port. From the few examples I have seen the bug only affects the y axis.

Image

I should be able to work on this issue and create a pull request.

Edit: This is a regression due to changes I made while supporting transforms on clipping paths. It should probably be fixed before the next release. I should have a PR up soon.

blayzen-w:
This PR fixes issue #487

While I was working on adding support for clipping paths I flattened the groups created in the convertShape function so the transform would be applied to a single group instead of multiple nested ones.

https://github.com/deeplook/svglib/blob/c82275981f50d8c2d3d6aa5d1118fd936d759047/src/svglib/svglib.py#L1938-L1944

However, just calling `applyTransformOnGroup` did not apply the transform in the correct order since matrix multiplication is non-commutative. This PR corrects that behavior.

### DIFF
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -1941,7 +1941,10 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
                 group.add(shape)
                 shape = group
 
-            self.applyTransformOnGroup(transform, shape)
+            # Matrix multiplication is non-commutative, need to apply in order
+            temp_group = Group()
+            self.applyTransformOnGroup(transform, temp_group)
+            shape.transform = mmult(temp_group.transform, shape.transform)
 
         return shape
 
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1801,6 +1801,31 @@ class TestEmbedded:
         assert image.width == 4
         assert image.height == 8
 
+    def test_image_with_transform(self):
+        """
+        This test will ensure that the image transform is correctly applied
+        to the image in a single group.
+        """
+        drawing = drawing_from_svg("""
+        <svg version="1.1" viewBox="0 0 50 50" width="50" height="50"
+            xmlns:xlink="http://www.w3.org/1999/xlink"
+            xmlns="http://www.w3.org/2000/svg"
+            xmlns:svg="http://www.w3.org/2000/svg">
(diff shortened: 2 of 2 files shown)


deeplook:
Thanks!

Issue #93, opened 2017-09-26
Embedded PNG images don't show up?


petri:
I get a "Adding box instead of image." warning for each embedded PNG image. The output PDF has no images. Any suggestions?

Here's the stanza for the embedded PNG images: 

`<image xlink:href="data:image/png;base64,iVBORw0KGgoAA.../>`

And the code:

```
from svglib.svglib import svg2rlg
from reportlab.graphics import renderPDF, renderPM

drawing = svg2rlg("input.svg")
renderPDF.drawToFile(drawing, "output.pdf")
```

deeplook:
Thanks for your report! I don't have a lot of time nowadays to look into this. But I'd suggest you look into the test suite for something similar and/or create a similar case with a sample PNG, either a non-copyrighted one or one you create dynamically using https://pypi.python.org/pypi/pypng.

deeplook:
Is this related to #94, maybe?

claudep:
@petri it would be nice to have an example file at hand.

jacobj:
+1 embedded png images don't seem to show up on both Python 3.5.4 and 3.6.2

deeplook:
Can anybody please provide reproducible code? Thanks! I'll add something around these lines in a *submitting bug reports* section in the README, soon...

petri:
The issue is easy enough to reproduce and there is no question about it being valid, imho. The fact that svglib warns "Adding box instead of image." regardless of what the image type is (PNG or other) suggests to me this is a known issue. Or if not, what does that warning mean if not that???

A proper test would have to check the internals of the resulting PDF to somehow detect whether the PNG is there. I don't have time to learn how to do that, just for a single test case that I have no certainty will result the issue getting fixed.

Having said that, here's a svg that can be used to confirm the issue (zipped because github does not support svg uploads): [png_in_svg.zip](https://github.com/deeplook/svglib/files/1358395/png_in_svg.zip)

I tried converting that and got the above-mentioned error.

petri:
To check the PDF, I am pretty sure pdfrw can do the trick: https://pypi.python.org/pypi/pdfrw

claudep:
Thanks for the sample file. I have a preliminary patch available here:
https://github.com/deeplook/svglib/compare/master...claudep:issue93
However, the image is reversed in the document. @deeplook I already struggled with shapes reversed in past patches, do you know what's wrong with coordinates?

deeplook:
[...]

PR #492, opened 2026-07-21
Don't truncate embedded <image> geometry to whole points (#490)

### LINKED ISSUE #490: Embedded <image> geometry is truncated to whole points, misaligning rasters from vector elements
# Embedded `<image>` geometry is truncated to whole points, misaligning rasters from vector elements

*Context: this surfaced while debugging why vector contours overlaid on a raster
image were slightly misregistered in a PDF built with ReportLab (via svglib).
AI-assisted debugging found the overlays were correct and the embedded image 
was being displaced. The minimal reproduction below isolates that to svglib's 
image-coordinate handling. I have a working monkeypatch and can create a PR,
but I'm unsure whether svglib's integer truncation is necessary or if this is
the correct approach. The following is LLM generated and reviewed by me.*

## Summary

`Svg2RlgShapeConverter.convertImage` builds the ReportLab `Image` from
**integer-truncated** coordinates while flipping the enclosing group by the
**un-truncated** value. As a result, an embedded raster `<image>` is placed up
to ~1 pt away from where the SVG specifies, so it no longer aligns with vector
elements (paths, rects, lines) given the same coordinates.

On `main`, `src/svglib/svglib.py`:

```python
image = Image(int(x), int(y + height), int(width), int(height), image)

group = Group(image)
group.translate(0, (y + height) * 2)   # <- uses the un-truncated y + height
group.scale(1, -1)
```

Because `x`, `y`, **`width` and `height`** are each truncated independently, the
error is not a constant offset — it grows across the image (the far edge is off
by more than the near edge), i.e. a small scale + shift rather than a pure
translation.

## Reproduction

```python
import base64, io
from PIL import Image as PILImage
from reportlab.graphics.shapes import Image, Rect, mmult, transformPoint

X, Y, W, H = 10.7, 20.7, 50.7, 50.7

buf = io.BytesIO(); PILImage.new("RGB", (8, 8), (0, 0, 0)).save(buf, "PNG")
href = "data:image/png;base64," + base64.b64encode(buf.getvalue()).decode()
svg = (
    '<svg xmlns="http://www.w3.org/2000/svg"'
    ' xmlns:xlink="http://www.w3.org/1999/xlink" width="100" height="100">'
    f'<image x="{X}" y="{Y}" width="{W}" height="{H}" xlink:href="{href}"/>'
    f'<rect x="{X}" y="{Y}" width="{W}" height="{H}" fill="none"/></svg>'
)

def corners(node, ctm):
    pts = [(node.x, node.y), (node.x + node.width, node.y + node.height)]
    return [transformPoint(ctm, p) for p in pts]

def walk(node, ctm, out):
    t = getattr(node, "transform", None)
    if t is not None:
        ctm = mmult(ctm, t)
    if isinstance(node, Image):
        out["image"] = corners(node, ctm)
    elif isinstance(node, Rect):
        out["rect"] = corners(node, ctm)
    for child in getattr(node, "contents", []):
        walk(child, ctm, out)

out = {}
walk(svg2rlg(io.BytesIO(svg.encode())), (1, 0, 0, 1, 0, 0), out)
(inl, ifr), (rnl, rfr) = out["image"], out["rect"]
print(f"near-edge x:  image={inl[0]:.3f}  rect={rnl[0]:.3f}  delta={inl[0]-rnl[0]:+.3f} pt")
print(f" far-edge x:  image={ifr[0]:.3f}  rect={rfr[0]:.3f}  delta={ifr[0]-rfr[0]:+.3f} pt")
```

Output (x is unaffected by the vertical flip, so it isolates the truncation):

```
near-edge x:  image=7.500  rect=8.025  delta=-0.525 pt
 far-edge x:  image=45.000  rect=46.050  delta=-1.050 pt
```

The rect (vector) is placed correctly; the image (raster) is offset, and the
offset differs at the two edges.

## Expected

The embedded image should be placed at the same coordinates as an equivalent
vector element — the two `delta` values above should be `0.000`.

## Environment

- svglib 2.0.2 (also present on `main`)
- reportlab 5.0.0
- Pillow 12.3.0
- Python 3.13

## Possible fix / question for maintainers

Dropping the four `int()` casts fixes it (verified locally — both deltas become
`0.000`):

```python
image = Image(x, y + height, width, height, image)
```

The truncation appears to date back to the original embedded-image support
(#93, 2017) and I couldn't find a rationale for it. Is the rounding intentional
— e.g. to snap rasters to whole pixels in the `renderPM` backend — or is it safe
to remove? Happy to open a PR (with a regression test) if you'd like.

jstorrs:
Fixes #490.

`Svg2RlgShapeConverter.convertImage` passed `x`, `y`, `width`, and `height`
through `int()` when building the ReportLab `Image`, while flipping the
enclosing group by the **un-truncated** value:

```python
image = Image(int(x), int(y + height), int(width), int(height), image)
group = Group(image)
group.translate(0, (y + height) * 2)   # un-truncated
group.scale(1, -1)
```

So a raster embedded at fractional coordinates landed up to ~1pt away from a
vector element (path, rect, line) given the same coordinates. Because
`width`/`height` were truncated too, the error was a small scale + shift rather
than a constant offset — it grew toward the far edge, so it couldn't be
cancelled by shifting the overlay.

As discussed in the issue, there's no apparent reason for the casts (they date
back to the original embedded-image support in #93, 2017), and dropping them
places the raster at full precision.

## Change

- Remove the four `int()` casts in `convertImage`.
- Add `TestEmbedded.test_image_fractional_geometry_not_truncated`, which places
  an `<image>` at fractional coordinates and asserts the resulting `Image`
  keeps `x`/`width`/`height` un-truncated (and `y == y + height`, matching the
  group re-flip). It fails on the old code (`10 == 10.7`) and passes now.

## Verification

`pytest tests/test_basic.py` → 89 passed, 1 skipped. The existing
`test_image_with_transform` and the `test_image_size_fallback_*` tests are
unaffected (their geometry is integral).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,12 @@
 
 ### Fixed
 
+- Embedded `<image>` geometry is no longer truncated to whole points. The
+  `x`, `y`, `width`, and `height` of an SVG `<image>` were each passed through
+  `int()`, so a raster placed at fractional coordinates landed up to ~1pt away
+  from a vector element with identical coordinates. Because width/height were
+  truncated too, the error grew across the image rather than being a constant
+  shift (issue #490).
 - Replaced `locale.getdefaultlocale()` (used for `<switch>` `systemLanguage`
   matching) with a small environment-variable lookup. `getdefaultlocale()` is
   deprecated and is **removed in Python 3.15**; the replacement keeps the same
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -2371,7 +2371,7 @@ class Svg2RlgShapeConverter(SvgShapeConverter):
                     width = image_width
                     height = image_height
 
-        image = Image(int(x), int(y + height), int(width), int(height), image)
+        image = Image(x, y + height, width, height, image)
 
         group = Group(image)
         group.translate(0, (y + height) * 2)
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1826,6 +1826,40 @@ class TestEmbedded:
         image_group = drawing.contents[0].contents[0]
         assert image_group.transform == (0.5, 0.0, 0.0, -0.5, 0.0, 100.0)
 
+    def test_image_fractional_geometry_not_truncated(self):
+        """Fractional <image> geometry must be preserved, not truncated to ints.
+
+        A raster placed at fractional coordinates should land at the same
+        position as a vector element with identical coordinates. Truncating
+        x/y/width/height to whole points offset the raster by up to ~1pt and,
+        because width/height were truncated too, the error grew across the
+        image rather than being a constant shift (see issue #490).
+        """
+        drawing = drawing_from_svg(
+            """
+            <?xml version="1.0"?>
+            <svg xmlns="http://www.w3.org/2000/svg"
+                 xmlns:xlink="http://www.w3.org/1999/xlink"
+                 width="100" height="100" version="1.1">
+
+                <image x="10.7" y="20.7" width="50.7" height="50.7"
+                    xlink:href="data:image/png;base64,iVBORw0KGgoAAAANS
+                UhEUgAAAAUAAAAFCAYAAACNbyblAAAACXBIWXMAAAPoAAAD6AG1e1JrAAAAGXRFWHRTb2Z0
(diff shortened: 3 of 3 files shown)


deeplook:
Thanks!

PR #493, opened 2026-07-25
fix: respect display:none set via the style attribute

### LINKED ISSUE #401: Element visibility in style attribute not respected
Hi, I love that a pure python svg library exists so thanks for that.
I'm trying to write a cross-platform sphinx script and requiring platform-fussy binaries like cairo and inkscape just to convert svg->pdf is a complexity I did not need.

--------------

There is inconsistency in visibility attributes and styles.

I noticed this whilst using an svg with layers (as groups) where some of them are disabled using `style="display:none"`.

This rendered as expected in cairosvg but included all the layers in svglib's output.

---------------

I'll include a simple svg to demonstrate the different flavours of the issue.

This renders a series of labels with corresponding black boxes which  should conditionally render to match the label.

```svg
<svg height="200" width="240" xmlns="http://www.w3.org/2000/svg">
    <text x="20" y="15">Always visible</text>
    <rect x="5" y="5" width="10" height="10" />

    <text x="20" y="35">display none attribute</text>
    <rect x="5" y="25" width="10" height="10" display="none" />

    <text x="20" y="55">visibility hidden attribute</text>
    <rect x="5" y="45" width="10" height="10" visibility="hidden" />

    <text x="20" y="75">display none style</text>
    <rect x="5" y="65" width="10" height="10" style="display:none" />

    <text x="20" y="95">visibility hidden style</text>
    <rect x="5" y="85" width="10" height="10" style="visibility:hidden" />
    
    <text x="20" y="115">visible group</text>
    <g>
        <rect x="5" y="105" width="10" height="10" />
    </g>

    <text x="20" y="135">display none group</text>
    <g display="none">
        <rect x="5" y="125" width="10" height="10" />
    </g>

    <text x="20" y="155">visibility hidden group</text>
    <g visibility="hidden">
        <rect x="5" y="145" width="10" height="10" />
    </g>

    <text x="20" y="175">display none group style</text>
    <g style="display: none;">
        <rect x="5" y="165" width="10" height="10" />
    </g>

    <text x="20" y="195">visibility hidden group style</text>
    <g style="visibility: hidden;">
        <rect x="5" y="185" width="10" height="10" />
    </g>
</svg>
```

When converted to pdf via cairosvg we get:
image

But with svglib we get:
image

And rendered directly in edge we get:
image

--------------

- when `display="none"` is set as an attribute, it always works as expected (which makes sense looking at https://github.com/deeplook/svglib/blob/main/svglib/svglib.py#L541 and https://github.com/deeplook/svglib/blob/main/svglib/svglib.py#L579 ).

- Interestingly it also seems to respect `display:none` in the style of a non-group element (which I can't seem to find a path to in svglib.py).

- `display:none` in the style of a group in not respected (which is the failure case I first saw)

- "visibility hidden" does not seem to be respected anywhere.

----------------

I'd love for this to match the cairosvg output, but right now we're just modifying our svgs to turn the display style into an attribute.

Sanjays2402:
Closes #401

`display` was only read as a presentation attribute, so `<g style="display:none">` and shapes carrying `style="display:none"` were still rendered while the `display="none"` form was correctly hidden. Both spellings are equivalent in SVG.

Added `SvgRenderer.get_display()`, which prefers a `display` declaration from the `style` attribute and falls back to the attribute, and used it in the group and shape branches. New test in `tests/test_basic.py` fails without the change and passes with it.

(`visibility:hidden` is a separate, still-unhandled part of that issue and is deliberately out of scope here.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@
 
 ### Fixed
 
+- `display: none` set through a `style` attribute is now respected. Only the
+  `display` presentation attribute was read, so `<g style="display:none">` and
+  `<rect style="display:none"/>` were still rendered (issue #401).
+
 - Embedded `<image>` geometry is no longer truncated to whole points. The
   `x`, `y`, `width`, and `height` of an SVG `<image>` were each passed through
   `int()`, so a raster placed at fractional coordinates landed up to ~1pt away
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -1192,7 +1192,7 @@ class SvgRenderer:
             item = self.renderA(node)
             parent.add(item)
         elif name == "g":
-            display = node.getAttribute("display")
+            display = self.get_display(node)
             item = self.renderG(node, clipping=clipping)
             if display != "none":
                 parent.add(item)
@@ -1238,7 +1238,7 @@ class SvgRenderer:
                     node._resolved_target = target
 
             item = self.shape_converter.convertShape(name, node, clipping)
-            display = node.getAttribute("display")
+            display = self.get_display(node)
             if item and display != "none":
                 fill_val = self.attrConverter.findAttr(node, "fill")
                 m = _GRADIENT_URL_RE.fullmatch(fill_val.strip()) if fill_val else None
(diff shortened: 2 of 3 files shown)


deeplook (approved):
Verified: checked out the branch, ran the new test plus the full `test_basic.py` suite (90 passed, 1 skipped). The fix is well-scoped — a dedicated `get_display()` helper (correctly non-inheriting, unlike `findAttr`) that prefers `style="display:..."" over the presentation attribute, applied at both the group and shape render sites. Known, disclosed limitations (CSS-class-driven display:none, !important) are reasonable to leave out of scope. LGTM.

PR #494, opened 2026-07-25
fix: look up unquoted font-family names containing spaces

### LINKED ISSUE #374: Error when svg file contains references to font with space in its name
When input SVG file refers to a font with a space in its name (e.g. "Cascadia Code"), calling svg2rlg results in `Unable to find a suitable font for 'font-family:Cascadia Code', weight:normal, style:normal` warning messages and the default font is used instead.

Specifically, what happens is that when the font name has a space in its name, convertFontFamily method of Svg2RlgAttributeConverter located in svglib.py splits the input font name into multiple strings using space character as a delimiter (line 406), and then feeds these split strings into find_font method of FontMap class located in fonts.py, resulting in all calls to find_font failing (because it gets "Cascadia" and "Code" as separate inputs during two separate calls).

There are plenty of fonts with a space character in their name, even in standard Windows distribution, so using space character as a separator is not warranted and breaks the functionality.

Sanjays2402:
Closes #374

`convertFontFamily` split the font-family attribute with `split_attr_list`, which replaces commas with spaces before `shlex.split`, so an unquoted family name like `Cascadia Code` became `Cascadia` and `Code` and never matched a font registered under its full name.

Splitting now separates only on commas, yielding the full name first and the individual words afterwards so the previous lookup still works as a fallback. New test in `tests/test_fonts.py` fails without the change (`Cascadia` instead of `Courier`) and passes with it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,13 @@
   from a vector element with identical coordinates. Because width/height were
   truncated too, the error grew across the image rather than being a constant
   shift (issue #490).
+- `font-family` values are now split per CSS instead of on whitespace. An
+  unquoted multi-word name such as `font-family: Cascadia Code` was split into
+  `Cascadia` and `Code`, so a font registered under its full name was never
+  found (issue #374). Commas inside a quoted name (`font-family: "Foo, Bar",
+  Arial`) no longer split the list either; that previously handed `shlex` an
+  unbalanced quote and raised `ValueError: No closing quotation`, aborting the
+  conversion of a valid value.
 - Replaced `locale.getdefaultlocale()` (used for `<switch>` `systemLanguage`
   matching) with a small environment-variable lookup. `getdefaultlocale()` is
   deprecated and is **removed in Python 3.15**; the replacement keeps the same
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -534,6 +534,39 @@ class Svg2RlgAttributeConverter(AttributeConverter):
         return shlex.split(attr.strip().replace(",", " "))
 
     @staticmethod
+    def _split_outside_quotes(attr: str) -> List[str]:
+        """Split on commas that are outside quoted strings, respecting escapes."""
+        segments = []
+        current = []
+        quote_char = None
+        escaped = False
+        for char in attr:
+            if escaped:
+                current.append(char)
+                escaped = False
+                continue
+            if char == "\\":
+                current.append(char)
+                escaped = True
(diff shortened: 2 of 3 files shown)


deeplook:
Thanks—this fixes the intended case: unquoted multi-word families such as `font-family: Cascadia Code` are now looked up as the complete name before falling back to individual words. The focused font tests pass.

One blocking edge case: `split_font_family_list()` uses `attr.split(",")`, which splits commas inside quoted CSS family names. For example:

```css
font-family: "Foo, Bar", Arial
```

causes `shlex.split()` to receive unmatched quote fragments and raises:

```text
ValueError: No closing quotation
```

This can abort SVG conversion for a valid `font-family` value.

Could we split on commas only when outside quoted strings (respecting escapes), and add a regression test for `"Foo, Bar", Arial`?

Sanjays2402:
Good catch, thanks. Fixed in 1686c3f: `split_font_family_list()` now walks the string and only breaks on commas that are outside a quoted section, tracking the active quote char and honouring backslash escapes, so `"Foo, Bar", Arial` yields `["Foo, Bar", "Arial"]` instead of feeding shlex unbalanced fragments.

Added `test_font_family_quoted_name_containing_comma` in tests/test_fonts.py covering your exact value with both double and single quotes, plus a `convertFontFamily()` assertion so the valid value resolves rather than aborting. I checked it is not vacuous: with the split reverted to `attr.split(",")` the test fails with `ValueError: No closing quotation`, and passes with the fix. Full tests/test_fonts.py is green (40 passed, 2 skipped).

deeplook:
I'd be happy to include this in the next release which is just around the corner once this is cleaned-up.

Sanjays2402:
Happy to get it in. Current state on `7cfb105`: all checks green, and I have just added the missing CHANGELOG entry under Unreleased/Fixed covering both halves (multi-word unquoted name lookup for #374, and commas inside a quoted name no longer feeding shlex an unbalanced quote).

The other loose end I noticed was the assertion in `test_font_family_quoted_name_containing_comma`: it originally pinned the resolved fallback name, which is environment-dependent and went red on the Ubuntu jobs, so `dfacd2c` reduced it to asserting a name comes back at all. The exact-list assertions on `split_font_family_list()` still cover the actual parsing.

If "cleaned-up" means something more specific, say the word and I will do it. The branch has four commits including a pre-commit.ci autofix, so I can squash to two (fix + test) if you would rather not carry that in history.

deeplook:
Thanks!

PR #497, opened 2026-07-30
Fix _GRADIENT_URL_RE for single and double quotes.

zbohm:
Fix `_GRADIENT_URL_RE` for all types of url value:

1. without quotation marks `url(#name)`
2. enclosed in single quotes `url('#name')`
3. enclosed in quotation marks `url("#name")`
4. spaces before and after the value `url( #name )`, `url( '#name' )`, `url( "#name" )`

### DIFF
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -867,7 +867,8 @@ class ExternalSVG:
 
 _BEZIER_KAPPA = 0.5523  # cubic bezier constant for circle approximation
 
-_GRADIENT_URL_RE = re.compile(r"url\(#([^)]+)\)")
+# Match values: url(#name), url('#name'), url("#name"), url( "#name" )
+_GRADIENT_URL_RE = re.compile(r"""url\(\s*(['"]?)#([^'")\s]+)\1\s*\)""")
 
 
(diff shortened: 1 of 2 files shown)


deeplook:
Nice fix — verified it handles all six quote/spacing forms and correctly rejects mismatched quotes. Two small suggestions:

1. Excluding whitespace from the capture lets you drop the `.rstrip()`:
   ```python
   _GRADIENT_URL_RE = re.compile(r"""url\(\s*(['"]?)(#[^'")\s]+)\1\s*\)""")
   ```
2. Could you add a quick regression test covering the quoted/spaced variants so this doesn't regress?

zbohm:
Edited according to comments. Tests added.

deeplook:
Thanks for the quick turnaround — the fix and the added tests do the job! Just the `pre-commit.ci` lint check to sort out:

1. `ruff format` — one file needs reformatting (e.g. `colors.Color(1,0,0,1)` → `colors.Color(1, 0, 0, 1)`)
2. E501 — `tests/test_basic.py:2074` is too long (109 > 88)

`make lint` auto-fixes both (or `uv run pre-commit run --all-files` to reproduce CI locally). Push and it should go green — thanks!

deeplook:
Thanks!

PR #499, opened 2026-07-31
fix: respect the visibility property on shapes

### LINKED ISSUE #359: visibility ignored using svg2rlg and renderPM.drawToFile
An svg element where the attribute "visibility" is set to "hidden" should not be plotted, but it is.

Small example:

```

svg_string = """
<ns0:svg xmlns:ns0="http://www.w3.org/2000/svg" xmlns:ns1="http://www.inkscape.org/namespaces/inkscape" xmlns:ns2="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd" width="210mm" height="297mm" viewBox="0 0 210 297" version="1.1" id="svg5" ns1:version="1.1 (c68e22c387, 2021-05-23)" ns2:docname="test.svg">
  <ns2:namedview id="namedview7" pagecolor="#ffffff" bordercolor="#666666" borderopacity="1.0" ns1:pageshadow="2" ns1:pageopacity="0.0" ns1:pagecheckerboard="0" ns1:document-units="mm" showgrid="false" ns1:zoom="0.67081229" ns1:cx="396.53418" ns1:cy="561.25984" ns1:window-width="1680" ns1:window-height="987" ns1:window-x="-8" ns1:window-y="-8" ns1:window-maximized="1" ns1:current-layer="layer1" />
  <ns0:defs id="defs2" />
  <ns0:g ns1:label="Ebene 1" ns1:groupmode="layer" id="layer1">
    <ns0:path style="fill:none;stroke:#FF0000;stroke-width:5.63426" d="M 93.91599,350.32155 V 204.97537 h 84.08396 84.08396 l -5.14971,13.04389 c -25.58743,64.81145 -44.58875,161.07403 -52.08331,263.85921 l -1.00544,13.78925 H 148.88072 93.91599 Z" id="path948" transform="scale(0.26458333)" ns1:label="LustigerPfad" visibility="hidden" />
  </ns0:g>
</ns0:svg>
"""

with open("my.svg", "w", encoding="utf-8") as file:
    file.write(svg_string)

drawing = svg2rlg("my.svg")

renderPM.drawToFile(drawing, "my.png", fmt="PNG")
```

Resulting image:

my

Sanjays2402:
Closes #359

`renderNode` only consulted `display`, so a shape with `visibility="hidden"` was still rendered. The value is now resolved with `findAttr`, which reads the `style` attribute and walks ancestors, so inheritance from a group and a `visibility="visible"` override on a child both behave correctly; `collapse` is treated as hidden too.

Regression test added in `tests/test_basic.py` next to the `display:none` test — it fails without the one-line change and passes with it.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@
 
 ### Fixed
 
+- The `visibility` property is now respected on shapes. Elements with
+  `visibility="hidden"` (or `collapse`), set directly or inherited from an
+  ancestor, were still rendered (issue #359).
+
 - `display: none` set through a `style` attribute is now respected. Only the
   `display` presentation attribute was read, so `<g style="display:none">` and
   `<rect style="display:none"/>` were still rendered (issue #401).
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -1295,7 +1295,8 @@ class SvgRenderer:
 
             item = self.shape_converter.convertShape(name, node, clipping)
             display = self.get_display(node)
-            if item and display != "none":
+            visibility = self.attrConverter.findAttr(node, "visibility")
+            if item and display != "none" and visibility not in ("hidden", "collapse"):
                 fill_val = self.attrConverter.findAttr(node, "fill")
                 m = _GRADIENT_URL_RE.fullmatch(fill_val.strip()) if fill_val else None
                 if m:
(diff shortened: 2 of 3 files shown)


deeplook:
Thanks for the fix and the thorough regression test — merging.

Issue #502, opened 2026-08-06
`font-size` gets PX_TO_PT applied twice, text renders 0.75x too small relative to the geometry


Zunhammer:
Same thing as #471, which was closed as expected behaviour. I don't think it is, and I can't reopen that one, so I'm filing it separately with the extra detail I dug up. The README describes the right rule, the code just doesn't implement it.

We hit this while evaluating an upgrade of a floor plan report generator from 1.5.1. The geometry came out fine, but every label was clearly too small against the drawing it belongs to, and I ended up in the same place @blayzen-w did.

The reason this is easy to get wrong (I got it wrong first time round too): the `Tf` operator really does say 12 for `font-size="16"`, exactly like the README says. But it sits inside a `cm` matrix that already carries the 0.75, so what actually gets rendered is 9pt.

Repro, only needs svglib + reportlab:

```python
import io, os, re, tempfile
from svglib.svglib import svg2rlg
from reportlab.graphics import renderPDF
from reportlab.pdfgen import canvas

svg = """<svg width="96" height="96" viewBox="0 0 96 96" xmlns="http://www.w3.org/2000/svg">
  <rect x="0" y="0" width="48" height="8" fill="black"/>
  <text x="10" y="50" font-size="16" font-family="Helvetica">HHHH</text>
</svg>"""

f = tempfile.NamedTemporaryFile("w", suffix=".svg", delete=False, encoding="utf-8")
f.write(svg)
f.close()
d = svg2rlg(f.name)
os.unlink(f.name)

buf = io.BytesIO()
c = canvas.Canvas(buf, pagesize=(d.width, d.height))
c.setPageCompression(0)   # so the content stream stays readable
renderPDF.draw(d, c, 0, 0)
c.save()

print("Drawing:", d.width, "x", d.height)
stream = max(re.findall(rb"stream\r?\n(.*?)endstream", buf.getvalue(), re.S), key=len)
for line in stream.decode("latin-1").splitlines():
    if line.strip():
        print(line.strip())
```

The bar is 48 units wide and the font-size is 16, so in any renderer the text is a third of the bar.

On 2.1.0 (trimmed to the interesting lines):

```
Drawing: 72.0 x 72.0
...
.75 0 0 -0.75 0 72 cm
n 0 0 48 8 re f
1 0 0 -1 0 0 cm
BT /F1 12 Tf 14.4 TL ET
BT 1 0 0 1 10 -50 Tm (HHHH) Tj T* ET
```

[...]

Issue #471, opened 2026-06-29
Text is not correctly sized


blayzen-w:
Take following contrived SVG that has two path elements that are aligned to a piece of text to mark its position and scale:
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg" width="800" height="800" viewBox="0 0 200 200" version="1.1" id="svg1">
  <g id="layer1">
    <text xml:space="preserve" style="font-size:100px;fill:#0000ff;font-family: 'Noto Sans';" x="50" y="100" id="text1">=</text>
    <rect style="fill:#ff0000;" id="rect1" width="10" height="10" x="55.5" y="61"/>
    <rect style="fill:#ff0000" id="rect1-9" width="10" height="10" y="41" x="55.5"/>
  </g>
</svg>
```

The browser browser renders out the following:

<img width="379" height="331" alt="Image" src="https://github.com/user-attachments/assets/b42bcc94-9ddd-4130-803b-54f7d955133b" />

However the generated pdf places the equals sign in a different location.

<img width="403" height="411" alt="Image" src="https://github.com/user-attachments/assets/749306d6-9c1e-4506-8ae1-7b2971be8884" />

I dug into pdf commands that ReportLab is generating and it seems like px and pt units are somehow getting mixed up. Using the same svg example I get the following commands.

```
....
BT 1 0 0 1 50 -100 Tm /F2+0 75 Tf 90 TL (=) Tj T* ET
....
n 55.5 61 10 10 re f
...
n 55.5 41 10 10 re f
...
```

The first line for drawing the text (BT) has a size of 75, at position 50, -100. However, the svg defines the text element as 100px with the position of 50px and 100px. I believe all of these values are considered the same user space unit, so the first value should be 100 if the position is in pixels.

When I switch the value from 100px to 100pt I get the following pdf code:

`BT 1 0 0 1 50 -100 Tm /F2+0 100 Tf 120 TL (=) Tj T* ET`

I thought this was just a issue with the PDF renderer, but it looks like text the issue also shows up when exporting back to a svg. Using the same example I get the following:

```
<text x="50.0" y="-100.0" style="font-family: Noto Sans; font-size: 75.0px;fill: rgb(0%,0%,100%); stroke:none;" transform="translate(0,-200) scale(1,-1)">=</text>
```

[...]

PR #503, opened 2026-08-09
fix: stop applying PX_TO_PT to font-size twice (#502)

Zunhammer:
The viewport group already carries the px-to-pt conversion, but font-size was converted again below it: applyStyleOnShape mapped it through convertLengthToPt, and convertText measured advance widths in points before adding them to x coordinates in user units. Text therefore rendered at 0.75x the size of the geometry around it, and multi-tspan runs were laid out in the wrong unit. Both now stay in user units, leaving the single conversion to the viewport transform.

Three tests in tests/test_units.py asserted the stored String.fontSize in points and so encoded the bug; they now assert the rendered size instead. The relational tests there and in test_basic.py are untouched and still pass. Adds regression coverage for the text-to-geometry ratio across seven root variants, for multi-tspan positioning and for per-character positioning; all six fail without the source change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- `font-size` no longer has `PX_TO_PT` applied twice. The viewport group
+  already carries the conversion, but `String.fontSize` and the advance widths
+  used to position text fragments were converted to points as well. Text
+  therefore rendered at 0.75x the size of the geometry in the same document,
+  and multi-`tspan` runs were laid out in the wrong unit (issue #502). Both now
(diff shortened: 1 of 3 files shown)


deeplook:
Thanks for the detailed writeup and repro — this is a genuine regression from the 2.0b1 unit rework, not expected behaviour like #471.

Verified independently: your repro script now emits `Tf 16` inside the `.75` viewport matrix (matching 1.5.1 output, vs. the buggy `Tf 12`), full test suite passes locally (188 passed, 5 skipped), and all 16 CI checks are green.

Fix is well-scoped, the rewritten tests correctly assert rendered size instead of raw `String.fontSize`, and the changelog entry is clear about the unit-semantics change. Merging — thanks for catching this.

PR #507, opened 2026-08-15
Ignore a malformed viewBox instead of raising TypeError

eeshsaxena:
An SVG with a viewBox that isn't four numbers crashes `svg2rlg`:

```python
import io
svg2rlg(io.StringIO('<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100"><rect width="5" height="5"/></svg>'))
# TypeError: Box.__new__() missing 1 required positional argument: 'height'
```

`get_box` does `values = [float(v) for v in view_box.split()]` then `return Box(*values)` without checking the count, so a viewBox with the wrong number of values raises `TypeError` (and a non-numeric one raises `ValueError`). Per the SVG spec an invalid viewBox is ignored, so I fall through to the existing width/height fallback when it doesn't parse to exactly four numbers. Valid viewBoxes are unaffected.

Added a test; it raises TypeError on `main` and passes with the change, and the rest of test_basic.py still passes. Found it by fuzzing `svg2rlg` with mutated SVG.

### DIFF
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -1757,8 +1757,14 @@ class SvgRenderer:
         if view_box:
             # viewBox defines a unitless user-coordinate space (SVG spec §8.1).
             # Parse as raw floats — never apply unit conversion here.
-            values = [float(v) for v in view_box.replace(",", " ").split()]
-            return Box(*values)
+            try:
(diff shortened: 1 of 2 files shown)


deeplook:
Thanks for the fix and tests!

PR #508, opened 2026-08-15
Don't raise IndexError on an unbalanced transform expression

eeshsaxena:
A transform attribute with an unmatched parenthesis crashes the converter:

```python
from svglib.svglib import Svg2RlgAttributeConverter
Svg2RlgAttributeConverter().convertTransform('scale(2')
# IndexError: list index out of range
```

`convertTransform` collects the indices of all `(`/`)` characters and walks them two at a time as open/close pairs via `range(0, len(brackets), 2)`, reading `brackets[i + 1]`. An unmatched parenthesis leaves an odd number of positions, so the last iteration indexes past the list. I step to `len(brackets) - 1` so a final unmatched bracket is skipped; the transform is then reported as unparsable and ignored, matching how other invalid transforms are handled. Valid transforms are unaffected.

Extended the existing `TestTransformAttrConverter` mapping with unbalanced cases; it raises IndexError on `main` and passes with the change. Found it by fuzzing `svg2rlg` with mutated SVG.

### DIFF
--- a/src/svglib/svglib.py
+++ b/src/svglib/svglib.py
@@ -480,7 +480,9 @@ class AttributeConverter:
         for i, lin in enumerate(line):
             if lin in "()":
                 brackets.append(i)
-        for i in range(0, len(brackets), 2):
+        # Step in pairs and stop before a final unmatched "(" or ")" so an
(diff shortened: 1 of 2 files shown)


deeplook:
Thanks for the fix and tests!

